PR #4, opened 2020-08-31
clock realization

nikita-parkhomenko:
[DEMO LINK](https://nikita-parkhomenko.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,32 @@
-import React from 'react';
+import React, { Component } from 'react';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends Component {
+  state = {
+    date: null,
+  };
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  componentDidMount() {
+    setInterval(() => {
+      const date = new Date();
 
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -1,25 +1,32 @@
...
+import React, { Component } from 'react';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends Component {
+  state = {
+    date: null,
```
nanccyy, line 7: good job 👍  but why do you set 'null' for the date?
It causes that the clock appears one second later after the page is downloaded.

nanccyy (approved):

src/App.js
vkryvytskyy, line 7: @nanccyy is right, make it new Date().toLocaleTimeString() by default

vkryvytskyy (approved):

src/App.js
```diff
@@ -1,25 +1,32 @@
...
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        <p>
+          Current time:
+          {' '}
+          {this.state.date}
```
vkryvytskyy, line 25: use destructuring
```suggestion
          {date}
```

PR #33, opened 2020-09-03
solution

Totowa95:
[DEMO LINK](https://Totowa95.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,17 @@
 import React from 'react';
 
+import Clock from './components/Clock';
 import './App.scss';
 
-const App = () => {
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -1,25 +1,17 @@
...
-      </p>
-    </div>
-  );
-};
+const App = () => (
+  <div className="App">
+    <h1>React clock</h1>
+    <p>
+      Current time:
+      {' '}
```
vkryvytskyy, line 11: mb this text is better to put inside Clock component?

vkryvytskyy (approved):

PR #58, opened 2020-10-21
Done

aionova:
[DEMO LINK](https://aionova.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,11 @@
 import React from 'react';
-
+import Clock from './components/Clock/Clock';
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+const App = () => (
+  <div className="App">
+    <Clock />
+  </div>
+);
 
 export default App;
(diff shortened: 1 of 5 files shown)


yazheviks (approved):
Класс!)

src/App.js
```diff
@@ -1,25 +1,57 @@
...
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    name: 0,
+  };
+
+  toggleVisible = () => {
```
aliceblunt8, line 11:
```suggestion
  toggleVisibility = () => {
```

src/components/Clock/Clock.js
```diff
@@ -0,0 +1,39 @@
...
+
+  componentDidMount() {
+    setInterval(() => {
+      this.setState({
+        time: (new Date()).toLocaleTimeString(),
+      });
+      // eslint-disable-next-line no-console
+      console.log(new Date().toLocaleTimeString());
+    }, 1000);
+  }
```
aliceblunt8, line 17: you also need to clear the timer when the Clock is removed from the DOM, cause now the clock is still ticking in the console event after the Clock component is destroyed
hint: use componentWillMount and clearInterval
https://reactjs.org/docs/state-and-lifecycle.html

src/components/Clock/Clock.js
```diff
@@ -0,0 +1,39 @@
...
+    return (
+      <div className="clock">
+        <h1>
+          {`Clock named ${this.props.name}`}
+        </h1>
+        <p>
+          {this.props.isClockVisible ? `Current time:
+          ${this.state.time}` : 'No clock available'}
+        </p>
+      </div>
```
aliceblunt8, line 29: it's better to destructure state and props

src/components/Clock/Clock.js
aliceblunt8: also, let's create a separate method for updating the time (we can call it 'tick')

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #119, opened 2021-01-15
Develop

vitalya2626:
[DEMO LINK](https://vitalya2626.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,24 +2,12 @@ import React from 'react';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+import Clock from './Clock';
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+const App = () => (
+  <div className="App">
+    <Clock />
+  </div>
+);
 
 export default App;
--- a/src/Clock.js
+++ b/src/Clock.js
@@ -1,0 +1,56 @@
+import React from 'react';
+
+class Clock extends React.Component {
+  state = {
+    date: new Date(),
+    name: 0,
+    visibility: false,
+  };
+
+  componentDidMount() {
+    setInterval(() => {
+      this.setState({ date: new Date() });
+      // eslint-disable-next-line no-console
+      console.log(this.state.date.toLocaleTimeString());
+    }, 1000);
+  }
+
+  changeName = () => {
+    this.setState((state) => {
+      const random = Math.floor(Math.random() * Math.floor(1000));
+
+      // eslint-disable-next-line no-console
+      console.log(`The Clock has been renamed from ${state.name} to ${random}`);
+
+      return { name: random };
+    });
+  }
+
+  isClockVisible = () => {
+    this.setState(state => ({ visibility: !state.visibility }));
+  };
+
+  render() {
+    const { date, name, visibility } = this.state;
+
+    return (
+      <div>
+        <h1>Clock</h1>
+        <h2>{`Name: ${name}`}</h2>
+        <h3>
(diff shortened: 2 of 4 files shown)


NikitaTudakov (approved):
Good luck)

src/Clock.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+
+class Clock extends React.Component {
+  state = {
+    date: new Date(),
+    name: 0,
+    visibility: false,
```
NikitaTudakov, line 7: maybe it could be better to set default visibility:true?

vitalya2626:
Не знаю, как остановить таймер в консоли

src/Clock.js
```diff
@@ -0,0 +1,56 @@
...
+      console.log(this.state.date.toLocaleTimeString());
+    }, 1000);
+  }
+
+  changeName = () => {
+    this.setState((state) => {
+      const random = Math.floor(Math.random() * Math.floor(1000));
+
+      // eslint-disable-next-line no-console
+      console.log(`The Clock has been renamed from ${state.name} to ${random}`);
```
NikitaTudakov, line 23: try to move console.log into conditions when visibilty true, so it could disappear, when clocks hidden

src/Clock.js
```diff
@@ -0,0 +1,55 @@
...
+
+      return { name: random };
+    });
+  }
+
+  isClockVisible = () => {
```
aliceblunt8, line 27:
```suggestion
  toggleClockVisibility = () => {
```

src/Clock.js
```diff
@@ -0,0 +1,55 @@
...
+
+  render() {
+    const { date, name, visibility } = this.state;
+
+    return (
+      <div>
+        <h1>Clock</h1>
+        <h2>{`Name: ${name}`}</h2>
+        <h3>
+          { visibility ? `Current time : ${date.toLocaleTimeString()}` : null}
```
aliceblunt8, line 39:
```suggestion
          { visibility && `Current time : ${date.toLocaleTimeString()}`}
```

src/Clock.js
```diff
@@ -0,0 +1,55 @@
...
+    date: new Date(),
+    name: 0,
+    visibility: false,
+  };
+
+  componentDidMount() {
+    setInterval(() => {
+      this.setState({ date: new Date() });
+    }, 1000);
+  }
```
aliceblunt8, line 14: save the result of setInterval into the variable and use in ComponentWillUnmount to clear the interval
https://reactjs.org/docs/state-and-lifecycle.html#adding-lifecycle-methods-to-a-class

src/Clock.js
aliceblunt8: move visibility variable and everything connected with it to the App and render the Clock component only if it's true

aliceblunt8 (changes requested):

src/App.js
```diff
@@ -1,25 +1,57 @@
...
+          </button>
+
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              if (isClockVisible) {
+                this.setState({
+                  clockName: Math.round(Math.random() * 1000),
+                });
```
varseniuk, line 47: You should also log into console messag, as this task requires:

> 3. Every time the name changes print a message in the console The Clock was renamed from oldName to newName.

src/App.js
```diff
@@ -1,25 +1,57 @@
...
+      <div className="App">
+        <h1>A Clock</h1>
+        {isClockVisible && <Clock name={clockName} />}
+        <div>
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              this.setState({ isClockVisible: true });
+            }}
```
varseniuk, line 25: It's a good practice to create separate methods(functions) for certain operations.
E.g. create method 
```
showClock = () => {
    this.setState({
      isClockVisible: true,
    });
  }
```

ant then just call it here:

`onClick={this.showClock}`

It will make the code more readable and allow you to reuse this function

varseniuk (changes requested):

varseniuk (approved):
👍

PR #127, opened 2021-01-15
Solution

Iryna713:
[DEMO LINK](https://Iryna713.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,63 @@
 import React from 'react';
+import Clock from './components/Clock';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    clockName: 0,
+  }
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  changeClockVisibility = ({ target }) => {
+    if (target.innerText === 'Show Clock') {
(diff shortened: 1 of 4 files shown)


5Mountains (approved):
Awesome work! 🥇 

You might want to consider a slight improvement in the code in the shape of removing `<React.Fragment>` because there is a return `<div>` anyway.

src/App.js
```diff
@@ -1,25 +1,63 @@
...
+    isClockVisible: true,
+    clockName: 0,
+  }
+
+  changeClockVisibility = ({ target }) => {
+    if (target.innerText === 'Show Clock') {
+      this.setState({ isClockVisible: true });
+    } else if (target.innerText === 'Hide Clock') {
+      this.setState({ isClockVisible: false });
+    }
```
vsm2k17, line 17: I think it would be easier just split changeClockVisibility into to function

src/App.scss
```diff
@@ -1 +1,35 @@
...
+
+  background-color: rgba(142, 208, 247, 0.25);
+  border-radius: 50%;
+  box-shadow: 0 0 350px 0 rgba(0,0,0,0.75);
+
+  &__btnBlock {
+    display: flex;
+    justify-content: center;
+
+    // margin-bottom: 50px;
```
vsm2k17, line 21: clean your code

vsm2k17 (approved):
good job, but read the comments

PR #128, opened 2021-01-15
solution

dtsehelnyk:
[DEMO LINK](https://dtsehelnyk.github.io/react_clock/)

### DIFF
--- a/src/components/Clock/Clock.js
+++ b/src/components/Clock/Clock.js
@@ -1,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class Clock extends React.Component {
+  state = {
+    time: new Date(),
+  };
+
+  timer = setInterval(() => {
+    this.setState({
+      time: new Date(),
+    });
+
+    // eslint-disable-next-line
+    console.log(new Date().toLocaleTimeString());
+  }, 1000);
+
+  componentDidUpdate(prevProps) {
+    if (this.props.timerId !== prevProps.timerId) {
+      // eslint-disable-next-line
+      console.log(`The timer was renamed to ${this.props.timerId}`);
+    }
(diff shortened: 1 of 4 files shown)


MarchD (approved):
👍

src/components/Clock/Clock.js
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  timer = setInterval(() => {
+    this.setState({
+      time: new Date(),
+    });
+
+    // eslint-disable-next-line
+    console.log(new Date().toLocaleTimeString());
+  }, 1000);
```
aliceblunt8, line 16: setInterval in ComponentDidMount
https://reactjs.org/docs/state-and-lifecycle.html#adding-lifecycle-methods-to-a-class

aliceblunt8 (changes requested):

src/components/Clock/Clock.js
varseniuk: Let's follow Alisa's advice

src/components/Clock/Clock.js
```diff
@@ -0,0 +1,45 @@
...
+
+  componentDidUpdate(prevProps) {
+    if (this.props.timerId !== prevProps.timerId) {
+      // eslint-disable-next-line
+      console.log(`
+        The timer was renamed from ${prevProps.timerId} to ${this.props.timerId}
+      `);
+    }
+  }
```
varseniuk, line 25: It's great that you have used componentDidUpdate, but for this task, you should only console.log the new name only when clicking the button (so you could easily put the 'console.log to your 'changeId' method)

varseniuk (changes requested):

varseniuk (approved):
👍

PR #130, opened 2021-01-15
soli

Ruslanwp:
[DEMO LINK](https://Ruslanwp.github.io/react_clock/)

### DIFF
--- a/src/Clock.jsx
+++ b/src/Clock.jsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// eslint-disable-next-line react/prefer-stateless-function
+export class Clock extends React.Component {
+  render() {
+    const { time, isClockVisible } = this.props;
+
+    if (isClockVisible) {
+      // eslint-disable-next-line no-console
+      console.log(time.toLocaleTimeString());
+    }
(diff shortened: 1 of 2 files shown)


src/Clock.jsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  componentDidUpdate(prevProps) {
+    const oldName = prevProps.clockName;
+    const curName = this.props.clockName;
+
+    if (oldName !== curName) {
+      console.log(curName);
+    }
+  }
```
varseniuk, line 24: It's great that you have used componentDidUpdate, but for this task, you should only console.log the new name only when clicking the button (what you are actually doing in your 'clockNewName'). No need in this 'componentDidUpdate'

src/Clock.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// eslint-disable-next-line react/prefer-stateless-function
```
varseniuk, line 4: Why did you disable es-lint here?

varseniuk (changes requested):

varseniuk (approved):

PR #132, opened 2021-01-15
Solution

sasha39612:
[DEMO LINK](https://sasha39612.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,108 @@
 import React from 'react';
-
+import PropTypes from 'prop-types';
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+function getRandomInt(max) {
+  return Math.floor(Math.random() * Math.floor(max));
+}
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    clockName: 0,
+  }
 
-  return (
-    <div className="App">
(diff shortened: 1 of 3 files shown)


5Mountains (approved):
Awesome solution! 🥇

src/component/Clock.js
```diff
@@ -0,0 +1,47 @@
...
+
+  componentWillUnmount() {
+    clearInterval(this.interval);
+  }
+
+  render() {
+    return (
+      <>
+        {` ${this.state.date} `}
+      </>
```
varseniuk, line 36: You don't need the fragment here. Let's use some other (more semantic) tag here

varseniuk (changes requested):

src/App.js
```diff
@@ -1,25 +1,74 @@
...
+          >
+            Hide Clock
+          </button>
+        </p>
+
+        <p>
+          <button
+            className="button"
+            type="button"
+            onClick={() => this.getRandomInt()}
```
varseniuk, line 64:
```suggestion
            onClick={this.getRandomInt}
```

src/component/Clock.js
```diff
@@ -0,0 +1,35 @@
...
+    clearInterval(this.interval);
+  }
+
+  render() {
+    const { date } = this.state;
+
+    return (
+      <>
+        { date }
+      </>
```
varseniuk, line 28: Don't use the fragment here. No need for it. Use more semantic tag ( 'p' or 'span')

varseniuk (approved):

PR #135, opened 2021-01-16
 react-clock

Nvalja:
[DEMO LINK](https://Nvalja.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,57 @@
 import React from 'react';
 
+import Clock from './components/Clock';
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+  };
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        <div className="App__clockface">
+          <p>
+            Current time:
+            <br />
+            {this.state.isClockVisible && <Clock name={this.state.clockName} />}
+          </p>
+        </div>
+        <div className="App__buttons">
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              this.setState({ isClockVisible: true });
+            }}
+          >
+            Show Clock
+          </button>
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              this.setState({ isClockVisible: false });
+            }}
+          >
+            Hide Clock
+          </button>
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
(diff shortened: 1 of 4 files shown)


VinnieJ-2k20 (approved):
Good work!

src/App.js
```diff
@@ -1,25 +1,57 @@
...
+            <br />
+            {this.state.isClockVisible && <Clock name={this.state.clockName} />}
+          </p>
+        </div>
+        <div className="App__buttons">
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              this.setState({ isClockVisible: true });
```
varseniuk, line 27: It's better to create separate methods in your components and just call them here onClick

src/App.js
```diff
@@ -1,25 +1,57 @@
...
+            }}
+          >
+            Show Clock
+          </button>
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              this.setState({ isClockVisible: false });
```
varseniuk, line 36: the same

src/App.js
```diff
@@ -1,25 +1,57 @@
...
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => {
+              this.setState({ clockName: Math.random() });
```
varseniuk, line 46: and the same

src/components/Clock.jsx
```diff
@@ -0,0 +1,45 @@
...
+       console.log(this.state.dateOut.toLocaleTimeString());
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.name !== this.props.name) {
+      // eslint-disable-next-line
+      console.log(`The Clock was renamed from ${prevProps.name} to ${this.props.name}`);
+    }
+  }
```
varseniuk, line 22: You just only need to console.log the new name. So you can make it directly when clicking the button

varseniuk (changes requested):

varseniuk (changes requested):

src/App.js
```diff
@@ -1,25 +1,67 @@
...
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  state = {
+    isClockVisible: 1,
+    clockName: 3.14,
+  };
 
+  switchClock = (buttonName) => {
+    this.setState({ isClockVisible: buttonName === 'Show Clock' ? 1 : false });
```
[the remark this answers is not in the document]
Nvalja: I've forgotten that can write like : this.setState({ isClockVisible: buttonName === 'Show Clock' }); 🤦‍♂️

 I can't use your method because I call 1 method to two buttons

varseniuk (approved):

PR #141, opened 2021-01-17
Clock

kerpichonka:
[DEMO LINK](https://kerpichonka.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.js
+++ b/src/components/Clock.js
@@ -1,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class Clock extends React.Component {
+  static propTypes = {
+    name: PropTypes.number.isRequired,
+  }
+
+  state= {
+    time: new Date(),
+  };
+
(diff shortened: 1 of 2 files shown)


Droshik (approved):

src/components/Clock.js
```diff
@@ -0,0 +1,46 @@
...
+  componentDidUpdate(prevProps) {
+    const { name } = this.props;
+
+    if (name === prevProps.name) {
+      return;
+    }
+
+    // eslint-disable-next-line
+    console.log(`The Clock was renamed from ${prevProps.name} to ${name}`);
+  }
```
varseniuk, line 33: It's good that you have used componentDidUpdate, but for this task, you just need only to log to the console new name after changing it. So you can simply put your console.log into your 'changeName' method

varseniuk (changes requested):

varseniuk (approved):

PR #143, opened 2021-01-18
Solution

andrew-zaycev:
[DEMO LINK](https://andrew-zaycev.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,50 @@
 import React from 'react';
+import { Clock } from './Components/Clock';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  state = {
+    clockVisible: true,
+    clockName: 1,
+  }
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  render() {
+    return (
+      <>
+        <div className="App">
+          <h1>React clock</h1>
+          <p>
+            Current time:
+            {' '}
+            {this.state.clockVisible
+            && <Clock clockName={this.state.clockName} /> }
+          </p>
+          <button
+            type="button"
+            onClick={() => this.setState({ clockVisible: false })}
+          >
+            Hide Clock
+          </button>
+          <button
+            type="button"
+            onClick={() => this.setState({ clockVisible: true })}
+          >
+            Show Clock
+          </button>
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+          <button
+            type="button"
+            onClick={() => {
+              this.setState({ clockName: Math.random() });
+            }}
+          >
+            Set rundom name
+          </button>
+        </div>
+      </>
+    );
+  }
+}
 
 export default App;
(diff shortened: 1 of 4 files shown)


sashalarin202 (approved):
great!

src/App.js
```diff
@@ -1,25 +1,50 @@
...
+          <h1>React clock</h1>
+          <p>
+            Current time:
+            {' '}
+            {this.state.clockVisible
+            && <Clock clockName={this.state.clockName} /> }
+          </p>
+          <button
+            type="button"
+            onClick={() => this.setState({ clockVisible: false })}
```
varseniuk, line 25: better to create separate methods and call them onClick

src/App.js
```diff
@@ -1,25 +1,50 @@
...
+          >
+            Hide Clock
+          </button>
+          <button
+            type="button"
+            onClick={() => this.setState({ clockVisible: true })}
```
varseniuk, line 31: same

src/App.js
```diff
@@ -1,25 +1,50 @@
...
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+          <button
+            type="button"
+            onClick={() => {
+              this.setState({ clockName: Math.random() });
```
varseniuk, line 39: same

src/Components/Clock/Clock.js
```diff
@@ -0,0 +1,39 @@
...
+      console.log(this.state.date);
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.clockName !== this.props.clockName) {
+      // eslint-disable-next-line no-console,max-len
+      console.log(`The Clock was renamed from ${prevProps.clockName} to ${this.props.clockName}`);
+    }
+  }
```
varseniuk, line 24: It's great that you have used componentDidUpdate, but for this task, there's no need for it, you should only console.log the new name only when clicking the button (so you could easily put the 'console.log' right directly after giving a new name)

varseniuk (changes requested):

src/Components/Clock/Clock.js
andrew-zaycev: What's wrong with componentDidUpdate? If I don't use this method then I have to store the old name somewhere and componentDidUpdate gets prevProps, I think this method is perfect for solving this problem.

src/App.js
andrew-zaycev: Done

src/App.js
andrew-zaycev: Done

src/App.js
andrew-zaycev: Done

src/Components/Clock/Clock.js
varseniuk: Of course, it's up to you. As I've told you earlier - it's great, that you have used it. But my point is that for simple logging into the console after renaming there would be more meaningful to do it right away where you changing the name. Now it's hard to understand that's renaming the clock is that thing that triggers logging into the console the new name. And if there be some other functionalities in this task, that would re-render component, every time this check will be called

varseniuk (approved):

src/Components/Clock/Clock.js
andrew-zaycev: I understood your idea, thanks for the explanation, I will take note

PR #146, opened 2021-01-18
Solution

ShyianO:
[DEMO LINK](https://ShyianO.github.io/react_clock/)

### DIFF
--- a/src/components/Clock/Clock.js
+++ b/src/components/Clock/Clock.js
@@ -1,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class Clock extends React.Component {
+  state = {
+    time: new Date(),
+  };
+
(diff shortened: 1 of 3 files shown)


src/components/Clock/Clock.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  componentDidUpdate({ name: oldName }) {
+    const { name } = this.props;
+
+    if (oldName !== name) {
+      // eslint-disable-next-line
+      console.log(`The Clock was renamed from ${oldName} to ${name}`);
+    }
+  }
```
varseniuk, line 27: You could just console.log it right directly when you changing the name and get rid of componentDidUpdate

varseniuk (approved):
Good, but read my comment

PR #150, opened 2021-01-19
Develop

WhiteRhin0:
- [DEMO LINK](https://WhiteRhin0.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,65 @@
 import React from 'react';
-
+import { Clock } from './components/Clock';
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+export class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    clockName: Math.random(),
+    clockNewName: Math.random(),
+  }
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  genereteRandomName = () => {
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -1,25 +1,65 @@
...
-    </div>
-  );
-};
-
-export default App;
+export class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    clockName: Math.random(),
+    clockNewName: Math.random(),
```
varseniuk, line 9: You don't need a new name at the beginning. Let it be 'null' initially

src/App.js
```diff
@@ -1,25 +1,65 @@
...
+  }
+
+  genereteRandomName = () => {
+    this.setState({ clockNewName: Math.random() });
+    // eslint-disable-next-line react/no-access-state-in-setstate
+    this.setState({ clockName: this.state.clockNewName });
```
varseniuk, line 15:
```suggestion
    // eslint-disable-next-line react/no-access-state-in-setstate
    this.setState({ clockName: Math.random() });
```
To get access to the previous state you should use functional setState. That's why you got linter error. don't disable es-lint

src/App.js
```diff
@@ -1,25 +1,65 @@
...
+
```
varseniuk, line 16: generate clockNewName here

varseniuk (changes requested):

varseniuk (approved):

PR #158, opened 2021-02-23
Solution

oleksandrorel:
- [DEMO LINK](https://oleksandrorel.github.io/react_clock/)
- Created App class component that has `isClockVisible`, `clockName` state properties , function `randomName` and method `render`
-  Function `randomName` generates value and pass it to the state and make respective output to console, click handler on button `Set random name` referenced to this function.
-  Method `render` returns respective layout and pass state property `clockName` to the `Clock` component.
- `Clock` component has state property `time`, `componentDidMount` method  set interval in 1 sec and pass to the state new value of time, `componentWillUnmount` method clear interval that was set, method `render` returns respective layout of time and generated random name.

### DIFF
--- a/src/components/Clock.js
+++ b/src/components/Clock.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class Clock extends React.Component {
+  state = {
+    time: new Date().toLocaleTimeString(),
+  };
+
+  componentDidMount() {
+    this.timerId = setInterval(() => {
+      this.setState({ time: new Date().toLocaleTimeString() });
+    }, 1000);
+  }
+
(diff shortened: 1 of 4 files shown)


src/components/Clock.js
```diff
@@ -0,0 +1,48 @@
...
+      <>
+        <p className="text">
+          Current time:
+          {' '}
+          { time }
+
+          {
+          // eslint-disable-next-line no-console
+          console.log(time)
+          }
```
vkryvytskyy, line 33: Почему ты делаешь console.log в render'e, а  не в `componentDidMount => setInterval`? перед/после записи в state можно и вывести в консоль

vkryvytskyy (approved):

PR #161, opened 2021-03-01
React Clock tasks

MrKartoshka:
Lifecycle methods tasks
Add buttons Show Clock and Hide Clock in the App changing isClockVisible variable in the App state.
The Clock should not be rendered when isClockVisible is false.
Change the Clock component to print the time not only on the page but also in the DevTools console.
Stop the timer when the Clock is hidden.
Check if it works correctly:
Hide and show the Clock several times and leave it visible

Prop update task
Add Set random name button to the App to set a random number as a clockName variable in the App state.
Pass the name to the <Clock name={this.state.clockName} />
Every time the name changes print a message in the console The Clock was renamed from oldName to newName.
Check if the message appears in the console between clock ticks

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,58 @@
 import React from 'react';
 
 import './App.scss';
+import { Clock } from './Clock';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    randomName: 'Gosha',
+  }
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  showHideClock = () => {
+    this.setState(prevState => ({ isClockVisible: !prevState.isClockVisible }
+    ));
+  };
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+  setRandomName = () => {
+    this.setState({ randomName: 'Alex' });
+  }
+
+  render() {
+    const { isClockVisible, randomName } = this.state;
+    const { showHideClock, setRandomName } = this;
+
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -1,25 +1,58 @@
...
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    randomName: 'Gosha',
```
vkryvytskyy, line 9: it's not random name

src/App.js
```diff
@@ -1,25 +1,58 @@
...
+  }
+
+  showHideClock = () => {
+    this.setState(prevState => ({ isClockVisible: !prevState.isClockVisible }
+    ));
+  };
+
+  setRandomName = () => {
+    this.setState({ randomName: 'Alex' });
```
vkryvytskyy, line 18: it's not random too, use Math.random

src/App.js
```diff
@@ -1,25 +1,58 @@
...
-    </div>
-  );
-};
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    randomName: 'Gosha',
+  }
+
+  showHideClock = () => {
```
vkryvytskyy, line 12:
```suggestion
  toggleClockVisibility = () => {
```

src/App.js
```diff
@@ -1,25 +1,58 @@
...
+            >
+              Hide Clock
+            </button>
+          )
+          : (
+            <button
+              type="button"
+              onClick={showHideClock}
+            >
+              Show Clock
```
vkryvytskyy, line 49:
```suggestion
            <button
              type="button"
              onClick={showHideClock}
            >
              {isClockVisible ? 'Hide Clock' : 'Show'}
            </button>
```
it's the same button, only text is different

vkryvytskyy (changes requested):
Where is a link to ghpages?

PR #204, opened 2021-04-04
Created Reack clock

haiduk2019:
[DEMO LINK](https://haiduk2019.github.io/react_clock/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,28 +1,33 @@
 # React Clock
+
 - Replace `<your_account>` with your Github username in the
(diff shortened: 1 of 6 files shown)


IhorDrul (changes requested):
Every time the `name` changes print a message in the console `The Clock was renamed from oldName to newName`.

в тебе немає цих пунктів зроблених

Serhii-Nikitin (approved):

danheim (approved):

PR #229, opened 2021-05-20
Solution v0.1

SkyPKWolf:
[DEMO LINK](https://SkyPKWolf.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,67 @@
 import React from 'react';
-
+import { Clock } from './components/Clock';
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    name: 1,
+  }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.name !== this.state.name) {
     // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+      console.log(`The Clock was renamed from ${prevState.name} to ${this.state.name}`);
+    }
+  }
 
-  return (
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,25 +1,67 @@
...
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+        <button
+          type="button"
+          onClick={() => {
```
victor-arutiunov, line 44: You can create function before rendering to make code more clear, but this solution is also good.

victor-arutiunov (commented):
Nice work!

VinnieJ-2k20 (changes requested):
Everything looks good, just missed one detail from task guidelines:
> Pass the name to the `<Clock name={this.state.clockName} />`

SkyPKWolf:
> Everything looks good, just missed one detail from task guidelines:
> 
> > Pass the name to the `<Clock name={this.state.clockName} />`

I didn't use this name so i forget to added this property

VinnieJ-2k20 (changes requested):
The name of the clock prints into console event if clock is not visible. It should not.
That's why you need to pass `name` to `Clock` and print it there.

src/App.js
```diff
@@ -1,25 +1,63 @@
...
+            this.setState({
+              isClockVisible: false,
+            });
+          }}
+        >
+          Hide Clock
+        </button>
+        <button
+          type="button"
+          onClick={this.renameClock}
```
varseniuk, line 54: you can create such function for hide/show clock as well

varseniuk (approved):
Good! 👍 
Left some comments

PR #247, opened 2021-05-21
solution

progmari:
[DEMO LINK](https://progmari.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,25 +1,62 @@
 import React from 'react';
+import { Clock } from './components/Clock';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+export class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    name: '',
+  }
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  hideClock = () => {
+    this.setState({
+      isClockVisible: false,
+    });
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+  showClock = () => {
+    this.setState({
+      isClockVisible: true,
+    });
+  }
 
-export default App;
+  changeName = () => {
+    this.setState({
+      name: (Math.random() * 100).toFixed(0),
+    });
+  }
+
+  render() {
+    const { name } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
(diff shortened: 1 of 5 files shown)


iryna-vys (approved):
Good job!
But in console Clock doesn't work correctly.
You can use componentWillUnmount() to fix that.

VinnieJ-2k20 (changes requested):
1. You need to set some number as initial name for the clock.
image
2. If I hide the clock, console keeps printing last registered time
image

src/App.js
```diff
@@ -1,25 +1,62 @@
...
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+export class App extends React.Component {
+  state = {
+    isClockVisible: true,
+    name: '0',
```
varseniuk, line 9:
```suggestion
    name: '0',
```
It's number or string? If you are using numbers (with Math.random), so no need to convert them into strings

src/App.js
```diff
@@ -1,25 +1,62 @@
...
+        <h1>React clock</h1>
+        {this.state.isClockVisible && <Clock name={name} />}
+        <div className="App__buttons">
+          <button
+            type="button"
+            onClick={this.showClock}
+          >
+            Show
+          </button>
+          {' '}
```
varseniuk, line 44:
```suggestion
          {' '}
```
You could use some css to create margins

src/components/Clock/Clock.js
```diff
@@ -0,0 +1,42 @@
...
+    intervalID: null,
+  }
+
+  componentDidMount() {
+    this.setState({
+      intervalID: setInterval(() => {
+        this.setState({ time: new Date() });
+        // eslint-disable-next-line
+        console.log(this.state.time.toLocaleTimeString())
+      }, 1000),
```
varseniuk, line 16:
```suggestion
  componentDidMount() {
    this.intervalID = setInterval(() => {
        this.setState({ time: new Date() });
        // eslint-disable-next-line
        console.log(this.state.time.toLocaleTimeString())
      }, 1000),
```
No need to write setInterval into the state

src/components/Clock/Clock.js
```diff
@@ -0,0 +1,42 @@
...
+
+  componentDidUpdate(prevProps) {
+    if (prevProps.name !== this.props.name) {
+      //  eslint-disable-next-line
+      console.log(`The Clock was renamed from ${prevProps.name} to ${this.props.name}`)
+    }
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.state.intervalID);
```
varseniuk, line 28:
```suggestion
    clearInterval(this.intervalID);
```
If you would use it without state

varseniuk (approved):
Good job! 👍 
Left some comments

PR #250, opened 2021-05-21
Solution-001

senkiv-oleh:
[DEMO LINK](https://senkiv-oleh.github.io/react_clock/)


src/App.js
```diff
@@ -1,25 +1,75 @@
...
+  }
+
+  render() {
+    const { isClockVisible, name } = this.state;
+
+    return (
+      <>
+        <div className="current-time">
+          {`Current time: `}
+          {isClockVisible ? <Clock /> : 'hide clock'}
```
varseniuk, line 29:
```suggestion
          {isClockVisible ? <Clock /> : 'hidden'}
```
It's not clear what 'hide clock' should mean. If you want to say that clock is hidden, it's better to write it like this

src/components/Clock.js
```diff
@@ -0,0 +1,31 @@
...
+export class Clock extends React.Component {
+  state = {
+    date: new Date(),
+  }
+
+  componentDidMount() {
+    this.state.timerID = setInterval(
+      () => this.setState({
+        date: new Date(),
+      }), 1000,
```
varseniuk, line 12:
```suggestion
    this.timerID = setInterval(
      () => this.setState({
        date: new Date(),
      }), 1000,
```
You should add smth to state using `setState` method (not just directly `this.state.timerId =...`, as you have written.
Also, there is no need to write setInterval to the stat. You may just use `this` for that purpose

src/components/Clock.js
```diff
@@ -0,0 +1,31 @@
...
+    );
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.state.timerID);
+  }
+
+  render() {
+    // eslint-disable-next-line no-console
+    console.log(this.state.date.toLocaleTimeString());
```
varseniuk, line 22: better to move this console.log to the `setInterval` method

src/data/names.js
```diff
@@ -0,0 +1,17 @@
+const names = [
+  'Spongebob',
```
varseniuk, line 2: Clever 👍

varseniuk (approved):
👍 
But please pay attention to my comments about state

PR #272, opened 2021-05-24
Solution

SergeyZub:
[DEMO LINK](https://SergeyZub.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,24 +2,99 @@ import React from 'react';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  timerID = null;
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      date: new Date(),
+      isClockVisible: true,
+      clockName: 'clock',
+    };
+  }
+
+  componentDidMount() {
+    this.startClock();
+  }
+
+  componentWillUnmount() {
+    this.stopClock();
+  }
+
+  startClock = () => {
+    if (this.timerID === null) {
+      this.timerID = setInterval(() => {
+        this.tick();
+
+        // eslint-disable-next-line
+        console.log(this.state.date.toLocaleTimeString());
+      }, 1000);
+    }
+  }
+
+  stopClock = () => {
+    clearInterval(this.timerID);
+    this.timerID = null;
+  }
+
+  tick = () => {
+    this.setState({
+      date: new Date(),
+    });
+  }
+
+  showClock = () => {
+    this.startClock();
+    this.setState({
+      isClockVisible: true,
+    });
+  }
+
+  hideClock = () => {
+    this.stopClock();
+    this.setState({
+      isClockVisible: false,
+    });
+  }
+
+  setRandomName = () => {
+    const oldName = this.state.clockName;
+
+    this.setState({
+      clockName: `${Math.random()}`,
+    },
     // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
 
 import './App.scss';
 
-const App = () => {
-  setInterval(() => {
-    const date = new Date();
+class App extends React.Component {
+  timerID = null;
```
varseniuk, line 6:
```suggestion
```
No need in this

src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
...
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      date: new Date(),
+      isClockVisible: true,
+      clockName: 'clock',
+    };
+  }
```
varseniuk, line 15:
```suggestion
    this.state = {
      date: new Date(),
      isClockVisible: true,
      clockName: 'clock',
    };
```
No need in constructor here. Just use `state` object

src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
...
+
+        // eslint-disable-next-line
+        console.log(this.state.date.toLocaleTimeString());
+      }, 1000);
+    }
+  }
+
+  stopClock = () => {
+    clearInterval(this.timerID);
+    this.timerID = null;
```
varseniuk, line 38:
```suggestion
```
redundant

src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
...
+
+  showClock = () => {
+    this.startClock();
+    this.setState({
+      isClockVisible: true,
+    });
+  }
+
+  hideClock = () => {
+    this.stopClock();
```
varseniuk, line 55:
```suggestion
```
redundant. If `isClockVisible` is false, then `componentWillUnmount` will work (which will mean that your interval will be cleared)

src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
...
+    this.setState({
+      isClockVisible: false,
+    });
+  }
+
+  setRandomName = () => {
+    const oldName = this.state.clockName;
```
varseniuk, line 62: Instead of creating this function you may check if prevProps.name === this.props.name, using `componentDidUpdate` and (prevProps)

src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      date: new Date(),
+      isClockVisible: true,
+      clockName: 'clock',
+    };
+  }
+
+  componentDidMount() {
```
varseniuk, line 17: Lifecycle methods should be applied to the `Clock` component (not to the `App`)

src/App.js
```diff
@@ -2,24 +2,99 @@ import React from 'react';
...
+
+  render() {
+    return (
+      <>
+        <div className="App">
+          <h1>React clock</h1>
+          <p>
+            Current time:
+            {' '}
+            {this.state.isClockVisible && this.state.date.toLocaleTimeString()}
```
varseniuk, line 79:
```suggestion
            {this.state.isClockVisible && this.state.date.toLocaleTimeString()}
```
Let's read the task description more carefully. You should create a separate component `Clock`. It should have `name` and all lifecycle methods.
Also it should have `state`, which will contain `date: new Date(),`

varseniuk (changes requested):
Your logic is almost correct. But let's follow the task description and divide the logic.
Let's create a separate component `Clock` and move some logic there

varseniuk (approved):
Good job! 👍

PR #276, opened 2021-05-24
solution

ivan-shpynda:
[DEMO LINK](https://ivan-shpynda.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.jsx
+++ b/src/components/Clock.jsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class Clock extends React.Component {
+  state = {
+    time: new Date().toLocaleTimeString(),
+  }
+
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


src/components/Clock.jsx
```diff
@@ -0,0 +1,42 @@
...
+  componentWillUnmount() {
+    clearInterval(this.intervalID);
+  }
+
+  render() {
+    return (
+      // eslint-disable-next-line
+      <> {console.log(this.state.time)}
+        {this.state.time}
+      </>
```
varseniuk, line 35:
```suggestion
      <span>{this.state.time}<span/>
```
Better to move `console.log` into the `setInterval` method.
also, fragment is not neccessary here. It would be better to use more semantic tags

varseniuk (approved):
Great! 👍 
But pay attention to my comment

PR #280, opened 2021-06-01
Develop

RomanDubovskyi:
### [DEMO LINK](https://romandubovskyi.github.io/react_clock/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,12 +8,19 @@ class App extends React.Component {
   state = {
     isClockVisible: false,
     time: new Date().toLocaleTimeString(),
+    clockName: 0,
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(_, prevState) {
     clearInterval(interval);
+    const { isClockVisible, clockName } = this.state;
 
-    if (this.state.isClockVisible) {
+    if (clockName !== prevState.clockName) {
+      // eslint-disable-next-line
+      console.log(`The Clock was renamed from ${prevState.clockName} to ${clockName}`)
+    }
+
+    if (isClockVisible) {
       interval = setInterval(() => {
         const date = new Date().toLocaleTimeString();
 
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -1,25 +1,71 @@
...
+    if (isClockVisible) {
+      interval = setInterval(() => {
+        const date = new Date().toLocaleTimeString();
+
+        // eslint-disable-next-line
+        console.log(date);
+        this.setState({ time: date });
+      }, 1000);
+    }
+  }
```
vkryvytskyy, line 32: ComponentDidUpdate should be in Clock component, where you check previous and current props name, if it changed - print it to console.

src/App.js
```diff
@@ -1,25 +1,71 @@
...
+
+class App extends React.Component {
+  state = {
+    isClockVisible: false,
+    time: new Date().toLocaleTimeString(),
+    clockName: 0,
+  }
+
+  componentDidUpdate(_, prevState) {
+    clearInterval(interval);
```
vkryvytskyy, line 15: It should be in componentWillUnmount in Clock component. So every time clock is getting hidden - interval removed

vkryvytskyy (changes requested):

src/App.js
RomanDubovskyi: staren'kyi, it's much better this way, clock ne lagaet 💪

src/clock/Clock.jsx
```diff
@@ -0,0 +1,49 @@
...
+      // eslint-disable-next-line
+      console.log(date);
+      this.setState({ time: date });
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps) {
+    const { name } = this.props;
+
+    if (name === prevProps.name) {
```
vkryvytskyy, line 24:
```suggestion
    if (name !== prevProps.name) {
```

https://user-images.githubusercontent.com/38065883/120624232-279ad200-c469-11eb-84d2-c3f5aff3e305.mov

Staren'kyi, clock wasn't renamed

vkryvytskyy (approved):

src/clock/Clock.jsx
RomanDubovskyi: f*ck, dummy mistake, fixed

PR #340, opened 2021-08-31
Develop

lera-yaskevych:
[DEMO LINK](https://lera-yaskevych.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,94 @@
-import React from 'react';
+import React, { Props } from 'react';
 import './App.scss';
 
-const App: React.FC = () => {
-  setInterval(() => {
-    const date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  time: Date;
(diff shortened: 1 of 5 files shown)


SashaRudenko (approved):

src/App.tsx
```diff
@@ -1,24 +1,91 @@
...
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean;
+  clockName: number;
 };
 
+class App extends React.Component<Props<State>, State> {
+  state = {
```
varseniuk, line 11:
```suggestion
  state: State = {
```

src/App.tsx
```diff
@@ -1,24 +1,91 @@
...
+    isClockVisible: false,
```
varseniuk, line 12:
```suggestion
    isClockVisible: true,
```
Clock should be visible by default so you can set it in the state to `true` and there's no need in componentDidMount here

src/components/Clock.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+class Clock extends React.Component<Props, State> {
+  state = {
+    time: new Date(),
+  };
+
+  componentDidMount() {
+    setInterval(() => {
+      this.state.time.toLocaleTimeString();
```
varseniuk, line 19: Would be great to clear this interval on `componentWillUnmount`

varseniuk (changes requested):
let's clear interval on component unmount

varseniuk (approved):
Great 🔥

PR #355, opened 2021-09-02
Clock

UlaDutchak:


src/App.tsx
```diff
@@ -1,24 +1,73 @@
...
+    clockName: 0,
+  };
+
+  randomClockName = () => {
+    this.setState({
+      clockName: Math.ceil(Math.random() * 100),
+    });
+  };
+
+  showClock = () => {
```
sasha-morozov, line 22: I would suggest creating one method instead of these two: showClock and hideClock.
Since you are changing the boolean value, as the opposite, you may use "!isClockVisible" in order to change the value of the variable.

sasha-morozov (approved):
Great Job!)

src/components/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  componentDidMount() {
+    return this.timer;
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    const oldName = prevProps.name;
+
+    return oldName !== this.props.name
+      ? console.log(`Clock was renamed from ${oldName} to ${this.props.name}`)
+      : console.log(this.state.time.toLocaleTimeString());
```
ArtemTopolskyi, line 29: Эта логика должны быть в setInterval колбеке

ArtemTopolskyi (changes requested):
Часы работают не по условию
image

ArtemTopolskyi (approved):

PR #356, opened 2021-09-02
Taras Holiuk's solution of Clock

TarasHoliuk:
[DEMO LINK](https://tarasholiuk.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,49 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock/Clock';
 
-const App: React.FC = () => {
-  setInterval(() => {
-    const date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean;
+  clockName: number | string;
 };
+
+class App extends React.Component<{}, State> {
+  state = {
+    isClockVisible: true,
+    clockName: 'Try to set my name!',
+  };
+
+  changeVisibility = () => this.setState(prevState => (
+    { isClockVisible: !prevState.isClockVisible }));
+
+  setName = () => this.setState({ clockName: Math.round(Math.random() * 100) });
+
+  render() {
+    const { isClockVisible, clockName } = this.state;
+
+    return (
+      <>
+        <button type="button" onClick={this.changeVisibility}>
+          {isClockVisible ? 'Hide' : 'Show'}
+        </button>
+
+        <div className="App">
+          <h1>React clock</h1>
+          {isClockVisible && <Clock />}
+        </div>
+
+        <button type="button" onClick={this.setName}>
+          Set random name
+        </button>
+
+        <p>
+          Name:
+          {' '}
+          {clockName}
+        </p>
+      </>
+    );
+  }
+}
 
 export default App;
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+interface State {
+  time: string;
+  intervalId?: NodeJS.Timeout | number;
+}
+
+export class Clock extends React.Component<{}, State> {
+  state = {
+    time: new Date().toLocaleTimeString(),
+    intervalId: 0,
+  };
+
+  componentDidMount() {
+    const intervalId = setInterval(() => {
+      this.setState({ time: new Date().toLocaleTimeString() });
+      // eslint-disable-next-line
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -1,24 +1,52 @@
...
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean;
+  clockName: number | string;
```
sasha-morozov, line 7: Imo, I would suggest just using number type. The string is redundant here since you're using Math.random() to generate the random name.

evgen1sider (approved):
nice work

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+
+type State = {
+  time: string;
+  intervalId?: NodeJS.Timeout | number;
```
sasha-morozov, line 5: You may use intervalID as a property of the class

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }, 1000);
+
+    this.setState({ intervalId });
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    const oldName = prevProps.clockName;
+
+    return oldName !== this.props.clockName
+    // eslint-disable-next-line
```
sasha-morozov, line 30: disable eslint for console.log for the entire file. The code would be much readable in this case

sasha-morozov (approved):
Good job!)

src/components/Clock/Clock.tsx
TarasHoliuk: 🔥

src/App.tsx
TarasHoliuk: I wanted to use `'Try to set my name!'` string like starting name but maybe it was not a good idea 👀

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  componentDidMount() {
+    return this.interval;
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    const oldName = prevProps.clockName;
+
+    return oldName !== this.props.clockName
+      ? console.log(`Clock was renamed from ${oldName} to ${this.props.clockName}`)
+      : console.log(this.state.time);
```
ArtemTopolskyi, line 30: этот консоль лог лучше вынести в коллбек setInterval

ArtemTopolskyi (changes requested):

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  time: string;
+};
+
+type Props = {
+  clockName: number | string;
+};
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    time: new Date().toLocaleTimeString(),
```
YuriiHoliuk, line 14: Almost always it is better to store raw data in the component state(redux store/database/localStorage etc.). Because in that you will have more possibilities how you can use stored data.

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  interval = setInterval(() => {
+    this.setState({ time: new Date().toLocaleTimeString() });
```
YuriiHoliuk, line 18: Using raw Date object in the state will allow you to call `toLocaleTimeString` just once(in the render)

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }, 1000);
+
+  componentDidMount() {
+    return this.interval;
+  }
```
YuriiHoliuk, line 23: useless

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+/* eslint-disable no-console */
+
+type State = {
+  time: string;
+};
+
+type Props = {
+  clockName: number | string;
```
YuriiHoliuk, line 9: Why do you need union type here?

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  render() {
+    const { clockName } = this.props;
+
+    return (
+      <div>
+        <p>
+          Current time:
+          {' '}
+          {this.state.time}
```
YuriiHoliuk, line 45: better to destruct state in the same way as props

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,52 @@
...
+  render() {
+    const { clockName } = this.props;
+    const { time } = this.state;
+
+    return (
+      <div>
+        <p>
+          Current time:
+          {' '}
+          {time.toLocaleTimeString()}
```
ArtemTopolskyi, line 43: use template string

ArtemTopolskyi (approved):

PR #360, opened 2021-09-03
React clock created

Andrii-Pinchuk:
[DEMO LINK](https://andrii-pinchuk.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,86 @@
 import React from 'react';
 import './App.scss';
 
-const App: React.FC = () => {
-  setInterval(() => {
-    const date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean;
+  clockName: number | string;
+  date: string;
 };
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    isClockVisible: true,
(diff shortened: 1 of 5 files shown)


kristof22122 (approved):
Good job!

src/App.tsx
```diff
@@ -1,24 +1,86 @@
...
 
+class App extends React.Component<{}, State> {
+  state: State = {
+    isClockVisible: true,
+    clockName: 'New',
+    date: (new Date()).toLocaleTimeString(),
+  };
+
+  componentDidMount() {
+    setInterval(() => {
```
varseniuk, line 18: Let's create the component `Clock` and handle all about the time inside it.
Also, you should clear the interval when the component disappears from the page (componentWillUnmount)

src/App.tsx
```diff
@@ -1,24 +1,86 @@
...
+        {this.state.isClockVisible
+          ? (
+            <>
+              <h1>{`React clock ${this.state.clockName}`}</h1>
+              <p>
+                Current time:
+                {' '}
+                {this.state.date}
+                {// eslint-disable-next-line no-console
+                  console.log(this.state.date)
```
varseniuk, line 57: This console.log should be inside setInterval

src/App.tsx
```diff
@@ -1,24 +1,86 @@
...
+      };
+    });
+  };
+
+  render() {
+    return (
+      <div className="App">
+
+        {this.state.isClockVisible
+          ? (
```
varseniuk, line 49:
```suggestion
        {this.state.isClockVisible && (
```
Use this syntax of conditional rendering instead of ternaries

varseniuk (changes requested):
Let's split the logic between the components and change the logic a bit

denys-cheporniuk (approved):

PR #376, opened 2021-09-06
solution

Ingvvvar:
[DEMO LINK](https://Ingvvvar.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+
+type State = {
+  time: string;
+};
+
+type Props = {
+  clockName: number;
+};
+
+export class Clock extends React.Component<Props, State> {
+  state: State = {
+    time: new Date().toLocaleTimeString(),
+  };
+
+  update = setInterval(() => {
+    this.setState({ time: new Date().toLocaleTimeString() });
+
+    // eslint-disable-next-line
+    console.log(this.state.time);
+  }, 1000);
+
+  componentDidMount() {
+    return this.update;
+  }
+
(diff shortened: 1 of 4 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  componentWillUnmount() {
+    clearInterval(this.update);
+  }
+
+  render() {
+    const { time } = this.state;
+
+    return (
+      <>
```
MaksymBeliever, line 35: You don't really need to use react fragment here

MaksymBeliever (approved):
Good job! But don't forget to use the bootstrap or other CSS libraries for React tasks

src/Clock.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <span>
+          Current time:
+          {' '}
+          {time}
```
ArtemTopolskyi, line 39: template string

src/Clock.tsx
```diff
@@ -0,0 +1,44 @@
...
+export class Clock extends React.Component<Props, State> {
+  state: State = {
+    time: new Date().toLocaleTimeString(),
+  };
+
+  update = setInterval(() => {
+    this.setState({ time: new Date().toLocaleTimeString() });
+
+    // eslint-disable-next-line
+    console.log(this.state.time);
```
ArtemTopolskyi, line 20: you can call toLocaleTimeString() only here and get rid of it in state and setState

ArtemTopolskyi (approved):
good job

PR #382, opened 2021-09-06
first try

zadorojniiarsen:
[DEMO LINK](https://zadorojniiarsen.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,83 @@
-import React from 'react';
+import React, { Props } from 'react';
+import Clock from './components/Clock/Clock';
 import './App.scss';
 
-const App: React.FC = () => {
-  window.setInterval(() => {
-    const date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean;
+  clockName: number;
 };
+
+class App extends React.Component<Props<State>, State> {
+  state = {
+    isClockVisible: true,
+    clockName: 0,
+  };
+
+  setRandomName = () => {
+    if (this.state.isClockVisible) {
+      this.setState(prevState => {
+        const newName = Math.trunc(Math.random() * 50 + 1);
+
+        // eslint-disable-next-line no-console
+        console.log(`The Clock was renamed from ${prevState.clockName} to ${newName}`);
+
+        return {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,24 +1,83 @@
-import React from 'react';
+import React, { Props } from 'react';
```
MaksymBeliever, line 1: Are you sure you need to import Props here?

src/App.tsx
```diff
@@ -1,24 +1,83 @@
...
+import Clock from './components/Clock/Clock';
```
MaksymBeliever, line 2: You can make that path better. Create index.tsx file for the Clock component

src/App.tsx
```diff
@@ -1,24 +1,83 @@
...
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean;
+  clockName: number;
 };
 
+class App extends React.Component<Props<State>, State> {
```
MaksymBeliever, line 10: I think you can use just `{}`, like that: `class App extends React.Component<{}, State>`

src/App.tsx
```diff
@@ -1,24 +1,83 @@
...
+    this.setState({ isClockVisible: true });
+  };
+
+  hideClock = () => {
+    this.setState({ isClockVisible: false });
+  };
+
+  render() {
+    const { isClockVisible, clockName } = this.state;
+    const { setRandomName, hideClock, showClock } = this;
```
MaksymBeliever, line 41: When you have more than 2 constants, will be better to write it in that way:
const {
setRandomName,
hideClock,
showClock,
} = this;

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,53 @@
...
+  componentWillUnmount() {
+    clearInterval(this.timer);
+  }
+
+  render() {
+    const { name } = this.props;
+    const { time } = this.state;
+
+    return (
+      <>
```
MaksymBeliever, line 34: You don't really need to use react fragment here. You already have `div` tag like wrapper

MaksymBeliever (approved):
Good job, but you can improve your code. I left some comments below

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  timer = setInterval(() => {
+    this.state.time.toLocaleTimeString();
+    this.setState({ time: new Date() });
+  }, 1000);
+
+  componentDidMount() {
+    return this.timer;
+  }
```
Denchepornyuk, line 23: componentDidMount тут нічого не робить, потрібно записати setInterval в середину в метод componentDidMount

Denchepornyuk (approved):

PR #432, opened 2021-10-13
Develop

Yashkindzhus:
[DEMO LINK](https://Yashkindzhus.github.io/react_clock/)


src/App.tsx
```diff
@@ -1,24 +1,52 @@
...
-};
+  state: AppState = {
+    isClockVisible: true,
+    clockName: Math.round(Math.random() * 10),
+  };
+
+  setClock = (isClockVisible: boolean) => {
+    this.setState({ isClockVisible });
+    this.buttonName = isClockVisible ? 'Show' : 'Hide';
+  };
```
DY-JS, line 22: It's great solution with button but there is something wrongin with this logic

src/App.tsx
```diff
@@ -1,24 +1,52 @@
...
+    this.setState({ clockName });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <>
+          <button
+            type="button"
+            onClick={() => this.setClock(!this.state.isClockVisible)}
```
DY-JS, line 34: According previous comment:
onClick={this.setClock}

src/App.tsx
```diff
@@ -1,24 +1,52 @@
...
-};
+  state: AppState = {
+    isClockVisible: true,
+    clockName: Math.round(Math.random() * 10),
+  };
+
+  setClock = (isClockVisible: boolean) => {
+    this.setState({ isClockVisible });
+    this.buttonName = this.state.isClockVisible ? 'Show' : 'Hide';
+  };
```
Anna-Tkachenko, line 22: Would be better to create function 'changeClockVisibility' without any arguments that set in state value '!isClockVisible'
And in render method use this value from state for displaying 'Show' or 'Hide' value for button

PR #483, opened 2021-12-07
Solution

olya-shyian:
[DEMO LINK](https://olya-shyian.github.io/react_clock/)


JamesSunderlanddd (approved):
I don't see any issues, looks good to me!

src/App.tsx
```diff
@@ -1,24 +1,64 @@
...
+
+        <button
+          className="App__button"
+          type="button"
+          onClick={() => {
+            if (isClockVisible) {
+              this.setState({
+                clockName: Math.round(Math.random() * 1000),
+              });
+            }
```
illiarerun, line 54: make a function for it, declaring whole function body in onClick handler is not a good idea

src/App.tsx
```diff
@@ -1,24 +1,64 @@
...
+        <h1>React clock</h1>
+
+        {isClockVisible && <Clock name={clockName} />}
+
+        <button
+          className="App__button"
+          type="button"
+          onClick={() => {
+            this.setState({ isClockVisible: true });
+          }}
```
illiarerun, line 31: Make a functions for all handlers

src/App.tsx
```diff
@@ -1,24 +1,64 @@
...
+        >
+          Show
+        </button>
+
+        <button
+          className="App__button"
+          type="button"
+          onClick={() => {
+            this.setState({ isClockVisible: false });
+          }}
```
illiarerun, line 41: This one too

src/App.tsx
```diff
@@ -1,24 +1,64 @@
...
+class App extends React.Component<{}, State> {
+  state = {
+    isClockVisible: true,
+    clockName: 0,
+  };
+
+  render() {
+    const { isClockVisible, clockName } = this.state;
+
+    return (
```
illiarerun, line 20: I still can see and click Hide and Set random name buttons when the clock is hidden, even though they are having no effect when clock is hidden, so either hide them or make disabled based on state

illiarerun (changes requested):
Good job 🔥  Let's work on fixes now:

src/App.tsx
```diff
@@ -1,24 +1,70 @@
...
+          >
+            Hide
+          </button>
+        )}
+
+        {isClockVisible && (
+          <button
+            className="App__button"
+            type="button"
+            onClick={() => this.createRandomName()}
```
illiarerun, line 60: You can use shorter version if you don't pass any params to the function
```suggestion
            onClick={this.createRandomName}
```

src/App.tsx
```diff
@@ -1,24 +1,70 @@
...
+          >
+            Set random name
+          </button>
```
illiarerun, line 63: You can combine these buttons under one condition by using fragment, or even use ternary and render either "Show" button or "Hide" and "Rename" buttons

illiarerun (approved):
GJ 🔥 Let's improve your solution a bit:

illiarerun (approved):

PR #485, opened 2021-12-08
completed react clock

i-stanko:


JamesSunderlanddd (approved):
Well Done!

src/App.tsx
```diff
@@ -1,24 +1,49 @@
...
+        <div className="buttons">
+          <button className="buttons__btn" type="button" onClick={this.showClock}>
+            Show Clock
+          </button>
+          <button className="buttons__btn" type="button" onClick={this.hideClock}>
+            Hide Clock
+          </button>
+          <button className="buttons__btn" type="button" onClick={this.renameClock}>
+            Rename Clock
+          </button>
```
illiarerun, line 42: I still can see and click Hide and Set random name buttons when the clock is hidden, even though they are having no effect when clock is hidden, so either hide them or make disabled based on state

illiarerun (changes requested):
Good job 🔥 Let's work on fixes now:

illiarerun (approved):
GJ 🔥

illiarerun (approved):

PR #486, opened 2021-12-08
Solution

msaikun:
-  [DEMO LINK](https://msaikun.github.io/react_clock/)


src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+  nameOfClock: string;
 };
 
+class App extends React.Component<{}, State> {
+  state: State = {
+    clockVisible: true,
+    nameOfClock: 'Time to approve my task',
+  };
+
+  randomClockNames = ['Time to study JS', 'Time to relax', 'Time to study React', 'Time to think about where to celebrate the New Year'];
```
dmtrhrytsak, line 16: Whoooa! A+ for this)

dmtrhrytsak (approved):
It's really perfect) Well, maybe, you could also disable buttons when the clock is hidden, but that's not that important, anyway) I would definitely look at this clock every day I get up🎀

src/App.tsx
illiarerun: Great idea, let's format it a bit
```suggestion
  randomClockNames = [
  	'Time to study JS',
  	'Time to relax',
  	'Time to study React', 
  	'Time to think about where to celebrate the New Year',
  ];
```

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+            Rename Clock
+          </button>
+
+          <button
+            className="App__button"
+            type="button"
+            onClick={this.makeClockHidden}
+          >
+            Hide Clock
+          </button>
```
illiarerun, line 68: Agree with Dima, it'd be great to disable buttons based on the clock state

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,42 @@
...
+export class Clock extends React.Component<Props, State> {
+  state: State = {
+    time: new Date().toLocaleTimeString(),
+  };
+
+  timerId = setInterval(() => {
+    this.setState({ time: new Date().toLocaleTimeString() });
+    // eslint-disable-next-line no-console
+    console.log(this.state.time);
+  }, 1000);
```
illiarerun, line 20: Let's do it in componentDidMount() method, to start countdown only after the first render

illiarerun (changes requested):
Great job 🔥 Let's work on details now:

illiarerun (approved):
GJ 🔥

PR #487, opened 2021-12-08
clock solution

skachkov23:
[DEMO](https://skachkov23.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,33 @@
 import React from 'react';
 import './App.scss';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+import Clock from './components/Clock';
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+export default class App extends React.Component {
+  state = {
+    isClockVisible: true,
+  };
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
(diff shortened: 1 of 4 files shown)


alex-kostomarov (approved):
LGTM!

src/App.tsx
```diff
@@ -1,24 +1,33 @@
...
-
-export default App;
+  render() {
+    return (
+      <div className="App">
+        <h1 className="App__title">React clock</h1>
+        <button
+          type="button"
+          className="App__button"
+          onClick={() => this.setState({ isClockVisible: true })}
```
illiarerun, line 18: move this logic to some method, e.g. showClock()

src/App.tsx
```diff
@@ -1,24 +1,33 @@
...
+        >
+          Show clock
+        </button>
+        <button
+          type="button"
+          className="App__button"
+          onClick={() => this.setState({ isClockVisible: false })}
```
illiarerun, line 25: This one too

src/App.tsx
```diff
@@ -1,24 +1,33 @@
...
+        >
+          Hide clock
+        </button>
```
illiarerun, line 28: if you want to improve your solution, you may change the disabled state of buttons if the clock is hidden/shown and also work on optional task

illiarerun (changes requested):
GJ 🔥  Few fixes left:

src/App.tsx
skachkov23: > if you want to improve your solution, you may change the disabled state of buttons if the clock is hidden/shown and also work on optional task

sorry, I tried, but always something goes wrong :(

illiarerun (approved):
GJ

PR #500, opened 2021-12-09
clock init

junglq:
[DEMO LINK](https://junglq.github.io/react_clock/)


alinaShchavinska:
your link took a day off, does not want to work😂🏖 Check it please

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  render() {
+    const { time } = this.state;
+
+    return (
+      <time>
+        Current time:
+        {' '}
+        {time.toLocaleTimeString()}
```
ArtemTopolskyi, line 33: use template string https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

ArtemTopolskyi (changes requested):
Change the Clock component to print the time not only on the page but also in the DevTools console.

src/App.tsx
```diff
@@ -1,24 +1,49 @@
...
+    isClockVisible: true,
+  };
+
+  showClock = () => {
+    this.setState({ isClockVisible: true });
+  };
+
+  hideClock = () => {
+    this.setState({ isClockVisible: false });
+  };
```
AlinaMatuschak, line 21: you can implement one method that changes visible value

src/App.tsx
```diff
@@ -1,24 +1,49 @@
...
+    const { isClockVisible } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        <button
+          type="button"
+          onClick={this.hideClock}
+          className="btn"
```
AlinaMatuschak, line 33: not enough informative class name

AlinaMatuschak (approved):

PR #505, opened 2021-12-09
solution

MagicBox15:
# React Clock
- [DEMO LINK](https://MagicBox15.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,78 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './component/Clock/Clock';
+import randomClockNames from './clockNames';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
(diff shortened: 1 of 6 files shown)


DudarDaryna (approved):
I like this clock 🔥🔥🔥

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
+    this.setState({
+      isVisible: true,
+    });
+  };
+
+  hideClock = () => {
+    this.setState({
+      isVisible: false,
+    });
+  };
```
AlinaMatuschak, line 33: you can create one method that set isClockVisible value

src/clockNames.ts
```diff
@@ -0,0 +1,21 @@
...
+  'Click! Click!',
+  'Click! Click! Click!',
+  'One more time and I\'ll tell your mom that\'s you ate all candies',
+  'What\'s wrong with you?',
+  'Okay...',
+  'Fine. Name the price.',
+  'Am I a joke to you?',
+  'I\'m calling the police!',
+  'Long story short...',
+  'HELP!!!',
```
AlinaMatuschak, line 16: 😂

AlinaMatuschak (approved):
🚀

src/App.tsx
MagicBox15: That makes sense 🤔  Thanks

PR #506, opened 2021-12-09
without styles

dlugash:
[DEMO LINK](https://dlugash.github.io/react_clock/)


aldmpr (approved):
good work!

src/App.tsx
```diff
@@ -1,24 +1,57 @@
...
+    this.setState({
+      isClockVisible: false,
+    });
+  };
+
+  showClock = () => {
+    this.setState({
+      isClockVisible: true,
+    });
+  };
```
AlinaMatuschak, line 24: you can create one method that set isClockVisible value

AlinaMatuschak (approved):
Great 🚀. But I believe that you could implement optional task 💪

PR #514, opened 2021-12-11
React clock

WorkAccCom:
[DEMO LINK](https://workacccom.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,84 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+class App extends React.Component {
+  state = {
+    timerId: undefined,
+    currentTime: new Date().toLocaleTimeString(),
+    isClockVisible: true,
+  };
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  componentDidMount() {
+    const id = setInterval(() => {
+      const time = new Date().toLocaleTimeString();
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+      // eslint-disable-next-line
+      console.log(time);
+
+      this.setState({
+        timerId: id,
+        currentTime: time,
+      });
+    }, 1000);
+  }
+
+  componentDidUpdate() {
+    if (this.state.isClockVisible && !this.state.timerId) {
+      const id = setInterval(() => {
+        const time = new Date().toLocaleTimeString();
+
+        // eslint-disable-next-line
+        console.log(time);
+
+        this.setState({
+          timerId: id,
+          currentTime: time,
+        });
+      }, 1000);
+    }
+  }
+
+  render(): React.ReactNode {
+    const {
+      currentTime,
+    } = this.state;
+
+    return (
+      <div className="App">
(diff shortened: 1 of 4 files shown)


VladDyrenkov (approved):
GJ

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+class App extends React.Component {
+  state = {
+    timerId: undefined,
```
ihor-jpeg, line 7: it's best to not explicitly initialize variables to `undefined`

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+class App extends React.Component {
```
ihor-jpeg, line 5: Use an interface to describe the state

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
+          <Clock
+            time={currentTime}
+            id={this.state.timerId}
+          />
+        )}
+
+        <div className="App__buttons">
+          <button
+            type="button"
+            onClick={() => this.setState({ isClockVisible: true })}
```
ihor-jpeg, line 56: do not make these functions anonymous. define them outside the render method and give them proper names

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
+            className="App__show-clock"
+          >
+            Show Clock
+          </button>
+
+          <button
+            type="button"
+            onClick={() => this.setState({
+              isClockVisible: false,
+              timerId: undefined,
```
ihor-jpeg, line 66: no explicit undefined

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,26 @@
...
+  time: string,
+  id?: number
+}
+
+// eslint-disable-next-line
+export class Clock extends React.Component<Props> {
+
+  componentWillUnmount() {
+    clearInterval(this.props.id);
+  }
```
ihor-jpeg, line 13: you start the clock in one component and you stop it in a different one. not very intuitive

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
+        timerId: id,
+        currentTime: time,
+      });
+    }, 1000);
+  };
+
+  render(): React.ReactNode {
+    const {
+      currentTime,
+    } = this.state;
```
ihor-jpeg, line 39: use this for other variables from your state

src/App.tsx
```diff
@@ -1,24 +1,78 @@
...
+
+      // eslint-disable-next-line
+      console.log(time);
+
+      this.setState({
+        timerId: id,
+        currentTime: time,
+      });
+    }, 1000);
+  };
```
ihor-jpeg, line 34: this seems like the logic of your clock. but it is in the App for some reason.
think about components as separate pieces. I want to take this clock and use it somewhere else. but in your code, some parts of it are in a different component

ihor-jpeg (changes requested):

src/App.tsx
WorkAccCom: @ihor-jpeg Ok. Could you give me a suggestion, please? If I'll use null instead of undefined, will it be ok?

src/App.tsx
WorkAccCom: Couldn't understand. Could you explain your thought, please?

src/App.tsx
ihor-jpeg: yes 👍

PR #534, opened 2022-01-18
v2

makssisua:
[DEMO LINK](https://makssisua.github.io/react_clock/)


src/react-app-env.d.ts
```diff
@@ -1 +1,5 @@
 /// <reference types="react-scripts" />
+
+type StateVisible = {
```
gnosis-frog, line 3: seems like you declared but not used it

gnosis-frog (approved):

PR #608, opened 2022-03-16
 react_clock

ievgen-volkov:


src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+type State = {
+  date: string,
+};
```
Bohdan-Maksymiuk, line 5:
```suggestion
type State = {
  date: string,
  timerId: NodeJS.Timer | undefined,
};
```

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export class Clock extends React.Component<{}, State> {
+  state: State = {
+    date: new Date().toLocaleTimeString(),
+  };
+
+  timerId: number | undefined;
```
Bohdan-Maksymiuk, line 12: You should pass it to the state

Bohdan-Maksymiuk: Undefined by default

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
```
Bohdan-Maksymiuk, line 15: and setState here

Bohdan-Maksymiuk (changes requested):
Good job. TimerId should be stored in state.

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+type State = {
+  date: string,
+  timerId: NodeJS.Timer | undefined,
```
aliceblunt8, line 5:
```suggestion
  timerId?: NodeJS.Timer,
```

aliceblunt8 (approved):

PR #627, opened 2022-03-29
Solution without Optional

Olexandr-Trubin-Koshyk:
[DEMO LINK](https://olexandr-trubin-koshyk.github.io/react_clock/)


addKiwi (changes requested):
Good start
Time should be printed in console every second
Lifecycle methods tasks: 4, 5.2

Olexandr-Trubin-Koshyk:
> Good start Time should be printed in console every second Lifecycle methods tasks: 4, 5.2

solaryasha (approved):

PR #646, opened 2022-05-03
Develop

OksanaTyshchenko:
[DEMO LINK](https://OksanaTyshchenko.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,57 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+export class App extends React.Component {
+  state = {
+    clockName: 0,
+    isClockVisible: false,
+  };
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
+  clockStateVisible = () => {
+    this.setState({
+      isClockVisible: true,
+    });
+  };
 
-  // eslint-disable-next-line
-  console.log(timerId);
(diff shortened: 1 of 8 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import './Clock.scss';
+
+type Props = {
+  name: number;
```
VETALEK, line 5: It is strange that `name` is not of type `string`

src/components/Clock.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.name !== this.props.name) {
+      // eslint-disable-next-line
+      console.log(`The Clock was renamed from ${prevProps.name} to ${this.props.name}`);
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.timerId) {
```
VETALEK, line 38: I am not sure, but i suppose that `timerId` may be `0`
```suggestion
    if (this.timerId !== null) {
```

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
 import './App.scss';
+import { Clock } from './components/Clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+export class App extends React.Component {
+  state = {
+    clockName: 0,
+    isClockVisible: false,
```
VETALEK, line 8: Why clock is unvisible by default?

VETALEK (changes requested):
All seems nice) But need a few little changes

danylo-onishchuk (approved):

PR #691, opened 2022-05-21
Solution_0.1

andriy-fesych:
[DEMO LINK](https://andriy-fesych.github.io/react_clock/)
Sometimes, time is logged in console twice with same timestamp. It occurs rarely, and when running on the slow machine, hiwever it happens, and I couldn`t find a reason why.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import './App.scss';
-import Clock from './components/Clock';
+import { Clock } from './components/Clock';
 
 type Props = {};
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,86 @@
 import React from 'react';
 import './App.scss';
+import Clock from './components/Clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+type Props = {};
 
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
-  // eslint-disable-next-line
-  console.log(timerId);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type State = {
+  isClockVisible: boolean,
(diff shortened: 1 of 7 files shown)


andriy-fesych:
Also I know I broke "no-access-state-in-setstate" rule. I`m going to fix it later, or may accept your ready-to-use suggestion 😄

andriy-fesych:
"no-access-state-in-setstate" fixed, just FYI

src/App.tsx
```diff
@@ -1,27 +1,113 @@
...
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
+type Props = {};
+
+type State = {
+  isClockVisible: boolean,
+  x: number,
```
vkryvytskyy, line 9: why do you need `x` in state? What does `x` means btw?

src/App.tsx
```diff
@@ -1,27 +1,113 @@
...
+  clockName: number,
 };
 
+class App extends React.Component <Props, State> {
+  state = {
+    x: 0,
+    isClockVisible: true,
+    clockName: 182,
```
vkryvytskyy, line 17: math.random()?

src/App.tsx
```diff
@@ -1,27 +1,113 @@
...
+  };
+
+  componentDidMount() {
+    setInterval(() => {
+      this.setState((prevState) => ({ x: prevState.x + 1 }));
+    }, 1000);
+  }
```
vkryvytskyy, line 24: still don't get why do you need this `x`

vkryvytskyy (changes requested):

src/App.tsx
andriy-fesych: renamed it for more clear definition. It`s not a part of the task, variable keeps the amount of time App component has been mounted. 
Do You want me to remove it? It`s not affecting clock component in any way.

src/App.tsx
andriy-fesych: Thank You. Revised.

src/App.tsx
andriy-fesych: As mentioned above, to have a time counter of an App being mounted.
Should I remove it?

vkryvytskyy: Would be nice to, but no strong opinion

PR #705, opened 2022-05-31
solution

SavMary:
[DEMO LINK](https://savmary.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,70 @@
 import React from 'react';
 import './App.scss';
+import Clock from './Components/Clock/Clock';
+import { names } from './NamesForClock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
-
-    // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
-  }, 1000);
-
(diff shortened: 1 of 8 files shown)


src/Components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+      clearInterval(this.timerId);
+    }
+  }
+
+  render() {
+    const { date } = this.state;
+
+    return (
+      <div className="clock">
+        <p className="clock-info">Current time:</p>
```
dellachaise, line 47: If we have scss file, then I think you should rename class with nesting: for example clock__info. And use nesting in scss file.

src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
+      clockName: names[Math.floor(Math.random() * names.length)],
+    });
+  };
+
+  render() {
+    const { isClockVisible, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1 className="app-name">{clockName}</h1>
```
dellachaise, line 40: If you have scss file, then I think you should rename classes with nesting: for example App__name. And use nesting in scss file.

dellachaise (approved):
Good job, great styles!

Please read only little comments.

PR #713, opened 2022-06-01
clock

kitaminary:


src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,42 @@
...
+      clearInterval(this.timerId);
+      // eslint-disable-next-line no-console
+      console.log('Clock was stopped');
+    }
+  }
+
+  render() {
+    return (
+      <span>
+        {this.state.time}
```
iwanttosmile4u, line 38: You can use destructuring before return and make your notation more simple leaving just {time}.

src/App.tsx
```diff
@@ -1,27 +1,54 @@
...
+
+  render() {
+    return (
+      <div className="content is-large block">
+        <h1>React clock</h1>
+        <span>
+          Current time:
+          {' '}
+        </span>
+        {this.state.isClockVisible && (
```
iwanttosmile4u, line 30: In this case also will be good to use destructuring.

iwanttosmile4u (commented):
Good job! I just have a small recomendation)

Anna-Tkachenko (approved):
Good job! 🔥

PR #724, opened 2022-06-02
solution

LukaszGryziewicz:
[DEMO LINK](https://LukaszGryziewicz.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,20 @@ import './App.scss';
 import { Clock } from './components/Clock';
 
 type State = {
+  clockName: number;
   isClockVisible: boolean
 };
 
 class App extends React.Component<{}, State> {
   state = {
+    clockName: 1,
     isClockVisible: true,
+  };
+
+  generateClockName = () => {
+    const randomNumber = Math.round(Math.random() * 1000);
+
+    this.setState({ clockName: randomNumber });
   };
 
   hideClock = () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,58 @@
...
+  };
+
+  render() {
+    const { clockName } = this.state;
+
+    return (
+      <div className="App">
+        <div className="App__wrapper">
+          {this.state.isClockVisible
+            && <Clock data-cy="time" name={clockName} />}
```
efdalyalcin, line 37: maybe you can take this to the bottom, that way the buttons will not jump

efdalyalcin (approved):
Nicely implemented. Only thing is maybe you can change the order of clock and buttons to prevent the jump although it is not related to functionality.

src/App.tsx
LukaszGryziewicz: done

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div className="clock">
+        <h2 className="clock__time">
+          Current time:
+          {time}
+        </h2>
+        <h3 className="clock__name">
+          Clock name:
+          {name}
+        </h3>
```
meforoma, line 55: let's maintain clock name visible when the counter is hidden
so we don't see counter
but still see the name

meforoma (changes requested):

meforoma (approved):

PR #725, opened 2022-06-02
First_commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_clock/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,24 @@
-// styles go here
+.App {
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.appConteiner {
+  display: flex;
+  flex-direction: column;
+  text-align: center;
+}
+
+.timer {
+  color: rgb(8, 0, 252);
+  font-size: 24px;
+}
+
+.clolck {
+  font-size: 32px;
+  margin-bottom: 10px;
+  border: 4px solid blue;
+  border-radius: 4px;
+}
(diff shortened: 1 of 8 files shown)


src/Components/Clock.tsx
```diff
@@ -0,0 +1,41 @@
...
+    if (this.timerId) {
+      clearInterval(this.timerId);
+    }
+  }
+
+  render() {
+    const { time } = this.state;
+
+    return (
+      <div className=" ">{ time }</div>
```
valerii-derkach, line 38: you don`t need empty class name here

src/App.scss
```diff
@@ -1 +1,24 @@
...
+  flex-direction: column;
+  text-align: center;
+}
+
+.timer {
+  color: rgb(8, 0, 252);
+  font-size: 24px;
+}
+
+.clolck {
```
valerii-derkach, line 19: make `clock, timer, appContainer` elements of  block `App `
use` &__clock` ... `&__timer`  (BEM)
it would be easier to identify styles you imported or mede yourself

valerii-derkach (changes requested):

src/Components/Clock.tsx
igoree88: Fixed.

src/App.scss
igoree88: You are right. Fixed.

Denchepornyuk (approved):

PR #728, opened 2022-06-06
task completed

andreisucman:
[DEMO LINK](https://andreisucman.github.io/react_clock/)

### DIFF
--- a/src/Components/Clock.tsx
+++ b/src/Components/Clock.tsx
@@ -1,0 +1,40 @@
+/* eslint-disable no-console */
+import React from 'react';
+import './Clock.scss';
+
+type Props = {};
+
+type State = {
+  date: string;
+};
+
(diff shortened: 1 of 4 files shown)


src/Components/Clock.tsx
```diff
@@ -0,0 +1,40 @@
...
+import React from 'react';
+import './Clock.scss';
+
+type Props = {};
+
+type State = {
+  date: string;
+};
+
+export class Clock extends React.Component<Props, State> {
```
59MrRobot, line 11: You can just replace `Props` with `{}` so you don't need the type Props declaration on line 5

59MrRobot (changes requested):
Good job, Andrei. Just one suggestion

meforoma (approved):

PR #731, opened 2022-06-07
Develop

iLebo-stack:
[DEMO LINK](https://iLebo-stack.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,63 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Components/Clock/Clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+interface State {
+  isClockVisible: boolean;
+  clockName: number;
+}
 
+class App extends React.Component<{}, State> {
+  state = {
+    isClockVisible: true,
+    clockName: 0,
+  };
+
+  componentDidUpdate() {
     // eslint-disable-next-line
-    console.log(date.toLocaleTimeString());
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
-      <p>
-        Current time:
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+    console.log(this.state.clockName);
+  }
```
LukaszGryziewicz, line 19: I think this should be in the Clock component, also right now it displays the name every time you use any of the buttons since you don't check if the name has changed

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+          >
+            Hide
+          </button>
+
+          <button
+            type="button"
+            onClick={() => this.generateRandomNumber()}
+          >
+            Set random name
+          </button>
```
LukaszGryziewicz, line 56: Nice, but you don't display the name anywhere

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        <p data-cy="time">
+          Current time:
+          {' '}
+          {isClockVisible && <Clock name={clockName} />}
+        </p>
```
LukaszGryziewicz, line 34: I think that this entire paragraph could be rendered from the clock component, instead of it being just text

LukaszGryziewicz (changes requested):
Nice, almost there, please check out my comments.

meforoma (approved):

PR #733, opened 2022-06-07
Solution

PRO-GRAM-MER:
[DEMO LINK](https://pro-gram-mer.github.io/react_clock/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,8 +1,7 @@
 // styles go here
 .Clock {
   display: flex;
-  
-  
+
   &__App {
     margin: 0 auto;
     align-items: center;
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,11 @@
 // styles go here
+.Clock {
+  display: flex;
+  
+  
+  &__App {
+    margin: 0 auto;
+    align-items: center;
+    text-align: center;
+  }
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-unescaped-entities */
 import React from 'react';
 import './App.scss';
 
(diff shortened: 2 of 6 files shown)


src/App.scss
```diff
@@ -1 +1,11 @@
...
+.Clock {
+  display: flex;
+  
+  
+  &__App {
+    margin: 0 auto;
+    align-items: center;
+    text-align: center;
+  }
+}
```
LukaszGryziewicz, line 11: missing empty last line

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
+    this.setState({ isClockVisible: true });
+  };
+
+  render() {
+    const { clockName } = this.state;
+
+    return (
+      <div className="Clock">
+        <div className="Clock__App">
+          {this.state.isClockVisible
```
LukaszGryziewicz, line 37: You could also use the destructed state here

LukaszGryziewicz (approved):
Good job! 👍

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
+  hideClock = () => {
+    this.setState({ isClockVisible: false });
+  };
+
+  showClock = () => {
+    this.setState({ isClockVisible: true });
+  };
+
+  render() {
+    const { clockName } = this.state;
```
meforoma, line 32:
```suggestion
    const { clockName, isClockVisible } = this.state;
```

src/App.tsx
meforoma, line 37: upvote
```suggestion
          {isClockVisible
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div className="clock">
+        <h2 className="clock__time">
+          Current time:
+          {time}
+        </h2>
+        <h3 className="clock__name">
+          Clock name:
+          {name}
+        </h3>
```
meforoma, line 55: let's keep clock name not hidden
so that you hide the time counter, but name is visible
and we can change name while counter is hidden

meforoma (changes requested):

meforoma (approved):

PR #736, opened 2022-06-07
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,72 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
+interface State {
+  clockName: string;
+  isClockVisible: boolean;
+}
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,75 @@
...
+
+class App extends React.Component<{}, State> {
+  state = {
+    isClockVisible: false,
+    clockName: '',
+  };
+
+  render() {
+    const { clockName } = this.state;
+    const { isClockVisible } = this.state;
```
59MrRobot, line 18: You can combine these lines

src/components/Clock.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+type Props = {};
```
59MrRobot, line 3: You do not need this line, you can just use an empty object on line 10

59MrRobot (changes requested):
Good job, just a few suggestions

src/App.tsx
MabuzaM: Sure thanx. Done.

PR #738, opened 2022-06-07
solution

ShapovalDenys:
https://shapovaldenys.github.io/react_clock/


vkryvytskyy (changes requested):
Also add `rename clock` button

vkryvytskyy (approved):

PR #740, opened 2022-06-07
React clock task

proznerka:
[DEMO LINK](https://proznerka.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,48 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './clock';
 
-const App: React.FC = () => {
-  const timerId: NodeJS.Timer = setInterval(() => {
-    const date: Date = new Date();
-
-    // eslint-disable-next-line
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
+  clockInvisible = () => (this.setState({ isClockVisible: false }));
 
-export default App;
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        <p>
+          Current time:
+          {' '}
```
LukaszGryziewicz, line 26: Hmm, does this have some use?

LukaszGryziewicz (approved):
Good job,  just remember that component names should begin with a capital letter  🥇

src/App.tsx
meforoma: it's just a way to place a space so linter is not angry 🙂

meforoma (approved):

PR #741, opened 2022-06-07
something is off

WojciechCzarnowski:
[DEMO LINK](https://WojciechCzarnowski.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,18 +1,20 @@
 import React from 'react';
 
 type State = {
-  time: Date;
+  time: string;
 };
 
-type Props = {};
-
-export class Clock extends React.Component<Props, State> {
+export class Clock extends React.Component<{}, State> {
   state = {
-    time: new Date(),
+    time: new Date().toLocaleTimeString(),
   };
 
+  timerId?: NodeJS.Timer;
+
   updateTimer = setInterval(() => {
-    this.setState({ time: new Date() });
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+type State = {
+  time: Date;
```
proznerka, line 4: `type State = {
  time: string
};`

src/components/Clock.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+type Props = {};
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    time: new Date(),
```
proznerka, line 11: time: new Date().toLocaleTimeString(),

src/components/Clock.tsx
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  updateTimer = setInterval(() => {
+    this.setState({ time: new Date() });
+  }, 1000);
```
proznerka, line 16: Between line 15 and 16 add `console.log(this.state.time);`

proznerka (changes requested):
For now, I'll let you with this, let's implement this, and see if it will work 😉

proznerka:
I think you miss the part with `this.timerId`

proznerka (approved):
Guess it works this time :)

meforoma (approved):

PR #742, opened 2022-06-08
process

Uktam19980416:


src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+            clockName: Math.round(Math.random() * 1000),
+          })}
+        >
+          Random Clock Name
+        </button>
+      </div>
+    );
+  }
+}
+// const timerId: NodeJS.Timer = setInterval(() => {
```
elkhayate, line 67: here also

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
-        {' '}
-        {/* Print the time here instead of DevTools */}
-      </p>
-    </div>
-  );
-};
+// import './App.scss';
+
+import { Clock } from './components/Clock';
+// import { Countdown } from './Countdown';
```
elkhayate, line 5: delete unnecessary comments

elkhayate (approved):
great job, i left some comments if you wanna do it

meforoma (changes requested):
please purge all commented code
and implement logging to devTools console:
```
- Change the Clock component to print the time not only on the page but also in the DevTools console.
- The time in the console should be printed only once each second
```

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+            <Clock from={clockName} />
+          )}
+        </p>
+        <button
+          type="button"
+          onClick={() => {
+            this.setState({
+              isClockVisible: true,
+            });
+          }}
```
ihor-jpeg, line 34: make a handler

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+          Start
+        </button>
+
+        <button
+          type="button"
+          onClick={() => {
+            this.setState({
+              isClockVisible: false,
+            });
+          }}
```
ihor-jpeg, line 45: make a handler in the class and simply use it here

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+        >
+          Stop
```
ihor-jpeg, line 47: it says stop, but it does not stop the timer. it only hides it

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface State {
+  time: string;
+}
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    time: new Date().toLocaleTimeString(),
+  };
+
+  timerId?: NodeJS.Timer;
```
ihor-jpeg, line 17: why not in the state?

ihor-jpeg (changes requested):

src/App.tsx
meforoma: clearInterval sets off on `componentWillUnmount`
```
componentWillUnmount() {
    if (this.timerId) {
      clearInterval(this.timerId);
    }
  }
```

PR #755, opened 2022-07-04
solution?

AndriiNychaiuk:
- [DEMO LINK](https://AndriiNychaiuk.github.io/react_clock/)

### DIFF
--- a/src/components/clock.tsx
+++ b/src/components/clock.tsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import './clock.scss';
+
+interface Props {
+  name: number,
+}
+
+export class Clock extends React.Component<Props> {
+  timerId: NodeJS.Timer = setInterval(() => {}, 0);
+
+  state = {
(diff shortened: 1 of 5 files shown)


src/components/clock.tsx
```diff
@@ -0,0 +1,40 @@
...
+    clearInterval(this.timerId);
+  }
+
+  render() {
+    return (
+      <p data-cy="time" className="clock">
+        Current time:
+        {' '}
+        <br />
+        {this.state.time}
```
jstmpelowycz, line 36: Not a good solution, use styles for such things

jstmpelowycz (changes requested):

src/components/clock.tsx
AndriiNychaiuk: I'm not sure what you mean, so I just made it one line to make it easier to read

jstmpelowycz (approved):

PR #790, opened 2022-07-12
solution

alex-burkivskyi:
[DEMO LINK](https://alex-burkivskyi.github.io/react_clock/)


juajum (approved):
good job)

mykola-koval (approved):
clock name should only be shown and changing when the clock itself is shown

PR #799, opened 2022-07-12
Solution

AntonYegorov:
[DEMO LINK](https://AntonYegorov.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,35 +1,49 @@
-import React from 'react';
-import './App.scss';
+import { Component } from 'react';
+import { Clock } from './components/Clock';
 
-function getRandomName(): string {
-  const value = Math.random().toString().slice(2, 6);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,35 +1,49 @@
...
-}
+class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: 'Clock №1',
+  };
+
+  componentDidMount() {
+    document.addEventListener('click', this.visibleClock);
+    document.addEventListener('contextmenu', this.hiddenClock);
```
Olexandr6, line 17: You should remove event listeners in `componentWillUmount`, as well

Olexandr6 (approved):
I can't see `Clock.tsx` file, but your solution works well

PR #800, opened 2022-07-12
solution

LAndrikevych:
[DEMO LINK](https://LAndrikevych.github.io/react_clock/)


mykola-koval (changes requested):
- clock name should not be visible and changed when the clock is hidden
 - only time (without date should be printed in console)

mykola-koval (approved):

PR #802, opened 2022-07-13
React clock

garikr:
[DEMO LINK](https://garikr.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,45 @@
+/* eslint-disable no-console */
+import React from 'react';
+
+type Props = {
+  clockName: string,
+};
+
+type State = {
+  date: Date;
+};
+
+export class Clock extends React.Component<Props, State> {
+  timerId = 0;
+
+  state = {
+    date: new Date(),
+  };
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    console.log(
+      `Old name: ${prevProps.clockName} / New name:${this.props.clockName}`,
+    );
+  }
+
+  componentWillUnmount() {
+    console.log('Remove timer', this.timerId);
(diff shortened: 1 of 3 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,48 @@
...
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    console.log(
+      `Old name: ${prevProps.clockName} / New name:${this.props.clockName}`,
+    );
```
Olexandr6, line 28: You don't need to print this msg every time the component updates, only if old name !== new name

src/Clock.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  state = {
+    date: new Date(),
+  };
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+    }, 1000);
+  }
```
Olexandr6, line 23: You missed a task requirement:
image

```suggestion
  componentDidMount() {
    this.timerId = window.setInterval(() => {
      this.setState({ date: new Date() });
    }, 1000);
  }
```

src/Clock.tsx
```diff
@@ -0,0 +1,48 @@
...
+type State = {
+  date: Date;
+};
+
+export class Clock extends React.Component<Props, State> {
+  timerId = 0;
+
+  state = {
+    date: new Date(),
+  };
```
Olexandr6, line 17: State should be first and then class fields and methods
```suggestion
   state = {
    date: new Date(),
   };
   
  timerId = 0;

  
```

Olexandr6 (changes requested):
Well done, but let's fix some things👇

src/Clock.tsx
garikr: Done!

Olexandr6 (approved):
Well done👍

PR #807, opened 2022-07-13
finished with Clock

omelika:
[DEMO LINK](https://omelika.github.io/react_clock/)


Olexandr6 (approved):
Can't see your `Clock.tsx` file, but looks like your solution works well

PR #808, opened 2022-07-13
solution

Victance:


Olexandr6 (approved):
Looks weird, cause you completed an old version of this task. And you didn't commit `Clock.tsx` file.
But you solution works as expected

PR #818, opened 2022-07-26
Solution

BudnikOleksii:
- [DEMO LINK](https://BudnikOleksii.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Math.random().toString().slice(2, 6);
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const { clockName } = this.props;
+    const { date } = this.state;
+
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
```
VETALEK, line 41:
```suggestion
        time is
```
you can write the plain text 😉

VETALEK, line 41:
```suggestion
        time is
```
you can write the plain text 😉

src/components/Clock.tsx
```diff
@@ -0,0 +1,49 @@
...
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+
+      // eslint-disable-next-line
+      console.log(this.state.date.toLocaleTimeString());
+    }, 1000);
+  }
+
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
```
VETALEK, line 28: i think you need to pass the `renamingTimerId` to `Clock`, so it won't change its name when removed 🤔

VETALEK (changes requested):
Good start 🚀

src/components/Clock.tsx
BudnikOleksii: > i think you need to pass the `renamingTimerId` to `Clock`, so it won't change its name when removed 🤔

Please, clarify, what do you mean? I have two different timers😅One in the Clock component that update time every second(timerId variable to store setInterval and clear it when we have no clock). Second in App to update clock name every 3300ms(the same name for variable)

src/App.tsx
```diff
@@ -1,39 +1,67 @@
...
-        </strong>
+  handleClick = () => {
+    this.timerId = this.setTimer();
+    this.setState({ hasClock: true });
+  };
 
-        {' time is '}
+  handleContextmenu = () => {
+    this.setState({ hasClock: false });
+    window.clearInterval(this.timerId);
```
solaryasha, line 53: It's overcomplicated. You don't need to clear and create new timer for calling `getRandomName`

solaryasha: Methods `handleContextmenu` and `handleClick` should change only visibility of the clock

solaryasha (changes requested):

solaryasha (approved):
Good job !

PR #829, opened 2022-08-03
solution

DarkMistyRoom:
[DEMO LINK](https://DarkMistyRoom.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Math.random().toString().slice(2, 6);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,48 @@
...
+    document.addEventListener('click', () => this.setState(
+      { hasClock: true },
+    ));
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    window.setInterval(
+      () => this.setState({ clockName: getRandomName() }), 3300,
+    );
```
Anna-Tkachenko, line 34: You should clear this interval on componentWillUnmount in cases, when App component can be destroyed

Anna-Tkachenko (changes requested):

alexandra-protyanova (approved):

PR #833, opened 2022-08-04
Solution

Sergii2705:
- [DEMO LINK](https://Sergii2705.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Math.random().toString().slice(2, 6);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,70 @@
...
+      // eslint-disable-next-line
+      console.log(`Renamed from ${prevName} to ${curName}`);
+    }
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount() {
+    window.clearInterval(this.timerIdRename);
```
VETALEK, line 43: you don't unmount your `App` component. Try to kill the renaming timer from the `Clock` component, or add a check, to not rename the clock if `state.hasClock` is `false` 😉

VETALEK (changes requested):
GJ 🔥 
commented

src/App.tsx
```diff
@@ -1,39 +1,77 @@
...
+      // eslint-disable-next-line
+      console.log(`Renamed from ${prevName} to ${curName}`);
+    }
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentWillUnmount() {
+    if (this.state.hasClock) {
+      window.clearInterval(this.timerIdRename);
```
VETALEK, line 42: try to `clearInterval` from the `Clock` component 😉 
your `App` component won't ever be unmounted

VETALEK (changes requested):
Almost 💪

src/App.tsx
Sergii2705: I don't understand why I have to do this?

VETALEK: you don't want clock to be renamed when it is removed 🤔

VETALEK (approved):
👍

PR #839, opened 2022-08-04
clock was implemented

nikita-biriukov:
[DEMO LINK](https://nikita-biriukov.github.io/react_clock/)


fledith (changes requested):
In the component App, u dint clear intervals and just call each time new
use method componentWillUnmount

VETALEK (approved):
GJ 🔥

PR #843, opened 2022-08-05
solution

DmitriyTeteryn:
[DEMO LINK](https://DmitriyTeteryn.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './component/clock';
+
 import './App.scss';
 
 function getRandomName(): string {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -7,33 +9,66 @@ function getRandomName(): string {
...
-  }, 1000);
+    if (prevName !== currentName) {
+      console.log(`Renamed from <${prevName}> to <${currentName}>`); // eslint-disable-line
+    }
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentWillUnmount() {
+    clearInterval(this.timerId);
```
PavloMarkov, line 41: componentWillUnmount() don't work for clearing Interval, because App doesn't close or reloaded. Think "clearInterval(this.timerId)" should be placed in "componentDidMount()" method when the state is changed to "hasClock: false"
image

src/App.tsx
```diff
@@ -7,33 +9,66 @@ function getRandomName(): string {
...
+    console.clear() // eslint-disable-line
+  }
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  handlerClock = (event: MouseEvent) => {
+    if (event.type === 'contextmenu') {
+      event.preventDefault();
```
PavloMarkov, line 48: " event.preventDefault()" in this task - it's like a shot in a knee))

PavloMarkov (changes requested):
great work! 🔥 🔥 🔥 🔥 
have some comments

Ihor-Karpyn (changes requested):
I still see renaming in the console when the clock is hidden. Also add `event.preventDefault` to avoid call context menu

src/App.tsx
```diff
@@ -7,33 +9,68 @@ function getRandomName(): string {
...
-  }, 1000);
+  componentDidUpdate(_: Readonly<{}>, prevState: Readonly<State>) {
+    const prevName = prevState.clockName;
+    const currentName = this.state.clockName;
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+    if (prevName !== currentName) {
+      console.log(`Renamed from <${prevName}> to <${currentName}>`); // eslint-disable-line
+    }
```
Ihor-Karpyn, line 41: Move to clock component

ivan-kadykalo (changes requested):
You still didn't fix an issue with renaming when the clock was stopped
image

ivan-kadykalo (approved):
🔥

PR #844, opened 2022-08-05
Solution

yaroslav-lizogub:
[DEMO LINK](https://yaroslav-lizogub.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,72 @@
+import React from 'react';
+
+type State = {
+  date: Date,
+};
+
+type Props = {
+  clockName: string,
+};
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    date: new Date(),
+  };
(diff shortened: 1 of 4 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,72 @@
...
+      // eslint-disable-next-line no-console
+      console.log(message);
+    }
+  }
+
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+  }
+
+  tick() {
```
rodiontseva, line 40: Method's name should start with a verb (so you could easily tell what your method actually do)

rodiontseva (changes requested):
good job!
need some fix:
- Method's name
- hide only clock

rodiontseva (approved):
great!

ivan-kadykalo (approved):
💪

PR #847, opened 2022-08-05
Implemented react clock components

illnino380:
[Demo link](https://illnino380.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,71 @@
 import React from 'react';
 import './App.scss';
 
+import { Clock } from './Clock';
+
 function getRandomName(): string {
   const value = Math.random().toString().slice(2, 6);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const date = new Date();
-  const clockName = getRandomName();
+interface State {
+  clockName: string,
+  hasClock: boolean,
+}
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    // ...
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,71 @@
...
+    }
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.handleClockTurnOff);
+    document.removeEventListener('click', this.handleClockTurnOn);
+    window.clearInterval(this.handleChangeClockName);
```
PavloMarkov, line 43: window.clearInterval(this.handleChangeClockName); - it wouldn't work here because App don't close the window or update. you should replace this clearance to componentDidUpdate because of changing the state
image

PavloMarkov (changes requested):
good work! 👍 
have 1 comment to solve

src/App.tsx
illnino380: Thank you! Really good point!

PavloMarkov (approved):
👍

Ihor-Karpyn (changes requested):
Split into two components. Timer should place in the other component; All subscriptions should be removed in componentWillUnmount. Also do not forgot to remove event listeners

https://www.geeksforgeeks.org/javascript-removeeventlistener-method-with-examples/

Ihor-Karpyn (approved):

PR #880, opened 2022-08-25
solution

Kulchevych:
[DEMO LINK](https://Kulchevych.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock/Clock';
 
 function getRandomName(): string {
   const value = Math.random().toString().slice(2, 6);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,48 @@
...
-  window.clearInterval(timerId);
+    window.setInterval(() => {
+      const clockName = getRandomName();
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+      this.setState({ clockName });
+    }, 3300);
+  }
```
sergii-nosachenko, line 35: I've found an issue here: when your component unmounts this function keeps working in background. You need to clearInterval() before it will unmount.
[React-clock.webm](https://user-images.githubusercontent.com/54940595/187459991-4135fe4e-985d-4296-89da-52dcd1cad153.webm)

sergii-nosachenko (changes requested):
Good work, as always)
I think you need to call clearInterval inside Clock component before it will unmount.

sergii-nosachenko (approved):

dellachaise (approved):
Well done 👍

PR #886, opened 2022-09-01
Solution

aleksandra-prognimak:
[DEMO LINK](https://aleksandra-prognimak.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,51 @@
...
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {this.state.date.toLocaleTimeString()}
+        </span>
+      </div>
+    );
+  }
```
tiserett, line 50: use destructuring here

tiserett (changes requested):
read checklist and remove eventListeners on componentWillUnmount

src/Clock.tsx
aleksandra-prognimak: Why, if it adds 2 extra lines of code?

tiserett: I feel like using destructuring where it's possible is a good tone of writing code, but if u like it this way, then u can leave it like this

tiserett (approved):
great🌚

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+    clockName: 'Clock-0',
+  };
+
+  componentDidMount() {
+    document.addEventListener('click', this.handleClickLeftKey);
+    document.addEventListener('contextmenu', this.handleClickRightKey);
+
+    window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
```
AlinaMatuschak, line 28: create separate variable for 3300.

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.handleClickLeftKey);
+    document.removeEventListener('click', this.handleClickRightKey);
```
AlinaMatuschak, line 33:
```suggestion
    document.removeEventListener('contextmenu', this.handleClickRightKey);
```

src/App.tsx
AlinaMatuschak, line 28: Should remove this interval on unmount

AlinaMatuschak (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,63 @@
...
-        <span className="Clock__time">
-          {today.toLocaleTimeString()}
-        </span>
+export class App extends React.Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  timerId = 0;
```
AlinaMatuschak, line 22: where do you set this id? I see just that you use it to clear interval, but your timerId always is 0 🤔

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #891, opened 2022-09-05
Solution

Dmytro-Matsehora:
[DEMO LINK](https://Dmytro-Matsehora.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
+
+type State = {
+  hasClock: boolean,
+  clockName: string,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,47 @@
...
+    });
+
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+
+    window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+  }
```
ibondrk, line 36: I don't found part of code where you remove this events. It's looks like they always would be on your page. So in App component should be `componentWillUnmount` method, this was part of the task.
image

ibondrk (changes requested):
Good start but you miss some points of this task, i think it should be easy for you to make it better)

ibondrk (approved):
Good job!

dellachaise (approved):
Great job 👍

PR #898, opened 2022-09-08
Solution

Jamal-Amoian:
[DEMO LINK](https://jamal-amoian.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+export class App extends React.Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: getRandomName(),
+  };
+
+  componentDidMount() {
+    document.addEventListener('contextmenu', () => {
+      this.setState({ hasClock: false });
+    });
```
NataliyaAn, line 25: in my opinion  you need  to add `event.preventDefault `according to conditions :
#hide the `Clock` on a right mouse click in the `document` (`contextmenu` event):
    ```
```js
    document.addEventListener('contextmenu', (event) => {
      event.preventDefault(); // not to show the context menu

      // put your code here
    });
```

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+  clockName: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: getRandomName(),
+  };
+
+  componentDidMount() {
```
NataliyaAn, line 22: I'm not sure, but as I understand it, if we start actions in componentDidMount, we must stop them in componentWillUnmount.
You start in componentDidMount
   document.addEventListener('contextmenu'......)
   document.addEventListener('click',.....)
   window.setInterval(()....)
So you need to stop in componentWillUnmount:
       document.removeEventListener('click', ....);
     document.removeEventListener('contextmenu', ....);
     window.clearInterval(.......);

NataliyaAn (approved):

src/App.tsx
Jamal-Amoian: It is a condition from the task
- `the Clock should be visible only when the hasClock is true;`

src/App.tsx
Jamal-Amoian: Thanks! Fixed it

> I'm not sure, but as I understand it, if we start actions in componentDidMount, we must stop them in componentWillUnmount. You start in componentDidMount document.addEventListener('contextmenu'......) document.addEventListener('click',.....) window.setInterval(()....) So you need to stop in componentWillUnmount: document.removeEventListener('click', ....); document.removeEventListener('contextmenu', ....); window.clearInterval(.......);

src/App.tsx
```diff
@@ -1,39 +1,80 @@
...
+
+//   // this code stops the timer
+//   window.clearInterval(timerId);
+
+//   return (
+//     <div className="App">
+//       <h1>React clock</h1>
+//     </div>
+//   );
+// };
```
denys-danyliuk, line 80: delete comments

denys-danyliuk (approved):
GJ!

PR #929, opened 2022-09-12
solution

Maxim-Gumeniuk:
[DEMO LINK](https://Maxim-Gumeniuk.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
 import './App.scss';
+import { Component } from 'react';
+import { Clock } from './component/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.isRight);
+    document.removeEventListener('click', this.isRight);
+    window.clearInterval(this.timerId);
+  }
```
anyakrainiuk, line 38: the timer is not stopped when the clock is hidden.
I think the timer should be stopped.
Знімок екрана 2022-09-13 о 12 24 29

Знімок екрана 2022-09-13 о 12 26 53

anyakrainiuk: and after clicking, when the clock starts to be visible something wrong is printed in console.info
Знімок екрана 2022-09-13 о 12 30 44

anyakrainiuk (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+    }, 3300);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.isRight);
+    document.removeEventListener('click', this.isLeft);
+    window.clearInterval(this.timerId);
+  }
+
+  isRight = (event: MouseEvent) => {
```
denys-danyliuk, line 39: 3 point in checklist

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
+
+  isLeft = () => {
```
denys-danyliuk, line 44: 3 point in checklist

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #935, opened 2022-09-13
React Clock Created

Reptiloid044:
[DEMO LINK](https://Reptiloid044.github.io/react_clock/)
Hi!


src/App.tsx
```diff
@@ -7,33 +8,60 @@ function getRandomName(): string {
...
-  let clockName = 'Clock-0';
+type State = {
+  hasClock: boolean,
+  clockName: string,
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: getRandomName(),
```
dellachaise, line 19: `Clock-0` should be default value.

src/App.tsx
```diff
@@ -7,33 +8,60 @@ function getRandomName(): string {
...
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    document.addEventListener('mouseup', this.handleClick);
```
dellachaise, line 25: Hide and show clock doen't work. Fix it.
Screenshot 2022-09-14 at 11 29 35

src/App.tsx
```diff
@@ -7,33 +8,60 @@ function getRandomName(): string {
...
+  componentWillUnmount() {
+    document.removeEventListener('mouseup', this.handleClick);
+    window.clearInterval(this.timerId);
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  handleClick = (event: MouseEvent) => {
+    if (event.button === 0) {
```
dellachaise, line 40: Why do you need this check? I think it's redundand.

src/App.tsx
```diff
@@ -7,33 +8,60 @@ function getRandomName(): string {
...
+      this.setState({
+        hasClock: true,
+      });
+    }
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+    if (event.button === 2) {
```
dellachaise, line 46: Why do you need this check? I think it's redundant.
And you should separate this logic to different function.

dellachaise (changes requested):
Good start 👍 

But need to add changes 👇

Reptiloid044:
Hi Daria!
I have removed the checker and added handleLeftlick / handleRightClick
Now it looks more understandable 👾
Thank you!

denys-danyliuk (changes requested):
doesn't work correctly, probably you forgot to do deploy🥲

Reptiloid044:
Hi !
Every time I have problems with deploy or npm test ; (
image
What I can do for normal work of npm test or deploy?
Now it is deployed =)

dellachaise (approved):
Well done 👍 

About problems with deploy - try in this cases:
1. Delete `node_modules` folder.
2. Run `npm install`.

PR #946, opened 2022-09-14
react clock done

Andrii-Dunets:
[DEMO LINK](https://Andrii-Dunets.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,57 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './clock';
```
plemiannik, line 3: where is clock? 
image

plemiannik (changes requested):

src/App.tsx
Andrii-Dunets: was some problem with deploy, now it's ok

src/clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+export class Clock extends React.Component<Props, State> {
+  state = {
+    date: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      const date = new Date();
```
plemiannik, line 22: why u create const date instead of use date in state?

src/clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      // eslint-disable-next-line no-console
+      console.info(date.toLocaleTimeString());
+
+      this.setState({ date });
+    }, 1000);
+  }
+
+  componentDidUpdate(props: Props) {
```
plemiannik, line 31: better 2 use some name with something like 'prev'

plemiannik (approved):

src/clock.tsx
denys-danyliuk: good point

src/clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+    if (props.name !== this.props.name) {
```
denys-danyliuk, line 32: 😵‍💫

denys-danyliuk (changes requested):
GJ! 
Pay attention to the comments

AndriiRut (approved):
Good job! 🚀

PR #967, opened 2022-09-29
Develop

ghost:
[DEMO LINK](https://kamal-tufekcic.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -16,9 +16,8 @@ export class App extends Component<{}, ClockState> {
   };
 
   componentDidMount() {
-    document.addEventListener('contextmenu', this.handleDocContextMenu);
-
     document.addEventListener('contextmenu', (event) => {
+      this.handleDocContextMenu();
       event.preventDefault();
     });
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,53 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './Clock';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


Quelaan1 (approved):

src/App.tsx
```diff
@@ -1,39 +1,52 @@
...
+
+  componentDidMount() {
+    document.addEventListener('contextmenu', this.handleDocContextMenu);
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+    });
```
ihor-jpeg, line 23: why not use one listener?

src/App.tsx
```diff
@@ -1,39 +1,52 @@
...
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+    document.addEventListener('click', this.handleDocClick);
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+    window.setInterval(() => {
```
ihor-jpeg, line 27: clear this interval

ihor-jpeg (changes requested):

src/App.tsx
kamal-tufekcic: I'm not following, why? That's the interval for the optional task, it logs clock renames

ihor-jpeg (approved):

PR #976, opened 2022-10-10
solution

bohdan-tron:
[DEMO LINK](https://bohdan-tron.github.io/react_clock/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4374,9 +4374,9 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "1.2.3",
-      "resolved": "https://registry.npmjs.org/@mate-academy/scripts/-/scripts-1.2.3.tgz",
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,39 +1,73 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/clock';
+
+type State = {
+  clockName: string,
+  visibleClock: boolean,
```
polosanya, line 7: For boolean values it is better to give names starting with is/are/can etc. so it's clear that value has boolean type.
image

src/components/clock.tsx
```diff
@@ -0,0 +1,54 @@
+import { Component } from 'react';
+
+type State = {
+  time: Date;
+};
+
+type Name = {
+  name: string;
+};
```
polosanya, line 9:
```suggestion
type Props = {
  name: string;
};
```
It is type for props of the component, not just name

polosanya (changes requested):
Nice start! Please check my comments to make your code cleaner)

polosanya (approved):
Well done!

PR #983, opened 2022-10-13
created clock with name updating

Apollonia5:
[DEMO LINK](https://Apollonia5.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,12 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './components/Clock';
 import './App.scss';
+
+type State = {
+  clockName: string;
+  hasClock: boolean;
+  timerId: number;
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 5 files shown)


nikitaovch99 (approved):

src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+  state: State = {
+    today: new Date(),
+    timerId: 0,
+  };
+
+  componentDidMount() {
+    this.state.timerId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      if (this.props.hasClock) {
```
OleksiiBlezniuk, line 23: will this component be rendered if this property will have `false` value?

src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+type State = {
+  today: Date,
+  timerId: number,
+};
+
+export class Clock extends Component<Props, State> {
+  state: State = {
+    today: new Date(),
```
OleksiiBlezniuk, line 15: why `today`? 🙂 strange name for time

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+    const { clockName, hasClock } = this.state;
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && <Clock clockName={clockName} hasClock={hasClock} />}
```
OleksiiBlezniuk, line 56: no need to pass `hasClock` here

src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+    timerId: 0,
```
OleksiiBlezniuk, line 16: no need to save this value in state, you can just save it as property of class

PR #985, opened 2022-10-13
Component clock has been created

iam-datsenko:
[DEMO LINK](https://iam-datsenko.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+/* eslint-disable no-console */
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


iPanda177 (approved):
GJ

src/App.tsx
```diff
@@ -1,39 +1,67 @@
...
+  hasClock: boolean,
+  clockName: string,
+  clockRandomNameId: number;
+};
+
+export class App extends Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+    clockRandomNameId: 0,
```
Anna-Tkachenko, line 22: let's not store this value in state
just add it as separate variable, as you added state for the App class

src/App.tsx
```diff
@@ -1,39 +1,67 @@
...
+    }, 3300);
+  }
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+  componentDidUpdate(_prevProps: {}, prevState: State) {
+    if (prevState.clockName !== this.state.clockName && this.state.hasClock) {
+      console.debug(`Renamed from ${prevState.clockName} to ${this.state.clockName}`);
```
Anna-Tkachenko, line 35: I don't see this message in console, check it please

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+type State = {
+  clockTime: Date;
+  timerId: number,
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    clockTime: new Date(),
+    timerId: 0,
```
Anna-Tkachenko, line 17: same here, let's store this variable outside of the state object

Anna-Tkachenko (changes requested):

src/App.tsx
Anna-Tkachenko: everything is great here

Anna-Tkachenko (approved):
Good job!

PR #990, opened 2022-10-14
solution was added

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,64 @@
...
+    hasClock: true,
+  };
+
+  updateClockNameId = 0;
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+  componentDidMount(): void {
```
Yeliseiev, line 24: Why you use types for 'component' methods?

Yeliseiev (approved):
Good job!

src/App.tsx
Maksym845: This is an autofill of VS code, I think it does not matter.

PR #991, opened 2022-10-14
Develop

grafit-off:
[DEMO LINK](https://grafit-off.github.io/react_clock/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Create a `Clock` class component that will update the time every second using a 
     Renamed from oldName to newName
     ```
 - to see `console.debug` messages enable the `verbose` level in DevTools console:
-    
+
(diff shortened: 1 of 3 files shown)


grafit-off:
> In your solution console debug appear after 4-5 console.info, but must be 3-4 according to the task, but according to code everything looks good https://prnt.sc/PmOkCFHFzQ7E
> 
> Good Job 🔥

Yep, already fixed

Anna-Tkachenko (changes requested):
Don't see info in console about clock name renaming, check it

grafit-off:
> Don't see info in console about clock name renaming, check it

Strange, because I have it. Maybe, you need to turn on Verbose in the console settings
image

Anna-Tkachenko (approved):
Sorry, everything is great ⭐️

PR #998, opened 2022-10-14
solution

myroslav-diiak:
[DEMO LINK](https://myroslav-diiak.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,8 @@
 import React from 'react';
+
 import './App.scss';
+
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
-          {clockName}
-        </strong>
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.contextMenu);
+    document.removeEventListener('click', this.leftClick);
+    window.clearInterval(this.nameTimerId);
+  }
 
-        {' time is '}
+  contextMenu = (event: MouseEvent) => {
```
illia-sosnytskyi, line 48: nice solution, but it`s better to name function by what exectly it does

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  leftClick = () => {
```
illia-sosnytskyi, line 53: same here

illia-sosnytskyi (approved):

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+  handleLeftClick = () => {
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {this.state.hasClock && (
```
Anna-Tkachenko, line 62: will be better to use destructuring for getting needed props from state object

Anna-Tkachenko (approved):
Good job! ⭐️

PR #999, opened 2022-10-14
App with Clock component was added

Yevheniya-Deryka:
[DEMO LINK](https://<your_account>.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,12 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+
+import { Clock } from './Components/Clock';
+
+type State = {
+  clockName: string,
+  hasClock: boolean,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


Dikey945:
Excellent

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+    }, 3300);
+
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
+
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
```
Anna-Tkachenko, line 37: you should reset all these created listeners and timer in componentWillUnmount method

Anna-Tkachenko (changes requested):

src/App.tsx
Yevheniya-Deryka: done!

Anna-Tkachenko (approved):
🔥

PR #1003, opened 2022-10-15
Solution

vladimir-dublya:
[DEMO LINK](https://vladimir-dublya.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,48 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
-function getRandomName(): string {
-  const value = Date.now().toString().slice(-4);
-
-  return `Clock-${value}`;
-}
-
-export const App: React.FC = () => {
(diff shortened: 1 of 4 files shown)


Andrew-Skoryk (changes requested):
1. [CODE KNOWLEDGE] - Check if you remove your event listener, when component will unmount from page.

nikitaovch99 (approved):
lgtm

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+      this.setState({ clockName: this.getRandomName() });
+    }, 3300);
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.RightClick);
+    document.removeEventListener('click', this.LeftClick);
+  }
```
Anna-Tkachenko, line 31: you should clear the interval in component will unmount method

Anna-Tkachenko (changes requested):

PR #1006, opened 2022-10-16
clock implemented

artem-musii:


DenysKleimenov (changes requested):
Remove your event listeners, when a component unmounts from the page, and don't forget about the interval that you've set.

DenysKleimenov (approved):

silchyn (approved):
Cool!

PR #1009, opened 2022-10-16
Develop

Proph7000:
[DEMO LINK](https://proph7000.github.io/react_clock/)

### DIFF
--- a/src/component/clock.tsx
+++ b/src/component/clock.tsx
@@ -1,0 +1,70 @@
+import { Component } from 'react';
+
+type State = {
+  today: string;
+};
+
+type Props = {
+  date: number;
+};
+
+function getToday(): string {
+  return (new Date()).toUTCString().slice(-12, -4);
+}
+
+function getRandomName(date: number): string {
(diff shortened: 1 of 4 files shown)


src/component/clock.tsx
```diff
@@ -0,0 +1,70 @@
...
+  timerIdClockName = 0;
+
+  timerIdToday = 0;
+
+  componentDidMount(): void {
+    this.timerIdToday = window.setInterval(() => {
+      this.setState({ today: getToday() });
+      // eslint-disable-next-line no-console
+      console.info(this.state.today);
+    }, 1000);
```
loralevitska, line 35: you need use interval for ranaming clockName and set it to `App` state `clockName` like on your example:

`const timerId = window.setInterval(() => {
    clockName = getRandomName();
  }, 3300);`
  
  
image

loralevitska (changes requested):

TarasHoliuk (changes requested):

Virass (approved):
Nice work 🔥

PR #1013, opened 2022-10-17
Create a `Clock` class component that will update the time every second

OleksandrGlulhovskiy:
[DEMO LINK](https://OleksandrGlulhovskiy.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
+
+type State = {
+  hasClock: boolean,
+  clockName: string,
+};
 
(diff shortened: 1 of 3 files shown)


SerhiiRomaniukWD (approved):
GJ!

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
+
+  handleLeftClick = () => {
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {this.state.hasClock && (
```
Anna-Tkachenko, line 61: will be better to use destructuring for getting needed props from the state object

Anna-Tkachenko (approved):
Good job!

PR #1021, opened 2022-10-18
react_clock

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,55 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './components/Clock';
 import './App.scss';
 
-function getRandomName(): string {
-  const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './components/Clock';
```
polosanya, line 2:
```suggestion
import { Clock } from './components/Clock.tsx';
```
image

polosanya (changes requested):
Please check my comment and fix the problem.
Also, please add Clock component to your commit, so we can check it too, because for now it is absent on GitHub

loralevitska (changes requested):
Hi, I see you haven't fixed previous comments, maybe you did not push your changes?
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

PR #1027, opened 2022-10-18
creating clock

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/react_clock/)


loralevitska (changes requested):
it looks good!
correct all the requirements for the task, if you don't understand any of the points, write in the slack chat and we will solve the problem together

image

polosanya (approved):
Good job!

PR #1031, opened 2022-10-19
Solution

rstarovoit:
[DEMO LINK](https://rstarovoit.github.io/react_clock/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4374,9 +4374,9 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "1.2.3",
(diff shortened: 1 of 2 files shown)


src/components/clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {today.toUTCString().slice(-12, -4)}
```
grafit-off, line 49: You have used `.toUTCString().slice(-12, -4)` more then one time, so you can create an separate function for that.

Andr1yK (approved):

PR #1032, opened 2022-10-19
clock implemented v.1

korniienkokarina:
[DEMO LINK](https://korniienkokarina.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,38 +19,40 @@ export class App extends React.Component<{}, State> {
     hasClock: true,
   };
 
-  timerId = 0;
+  nameChangeTimer = 0;
 
   componentDidMount() {
-    document.addEventListener('contextmenu', this.rightClick);
-    document.addEventListener('click', this.leftClick);
+    document.addEventListener('contextmenu', this.handleRightClick);
+    document.addEventListener('click', this.handleLeftClick);
 
-    this.timerId = window.setInterval(() => {
+    this.nameChangeTimer = window.setInterval(() => {
       this.setState({ clockName: getRandomName() });
     }, 3300);
   }
 
   componentWillUnmount() {
-    document.removeEventListener('contextmenu', this.rightClick);
-    document.removeEventListener('click', this.leftClick);
+    document.removeEventListener('contextmenu', this.handleRightClick);
+    document.removeEventListener('click', this.handleLeftClick);
 
-    window.clearInterval(this.timerId);
+    window.clearInterval(this.nameChangeTimer);
   }
 
-  rightClick = (event: MouseEvent) => {
+  handleRightClick = (event: MouseEvent) => {
     event.preventDefault();
     this.setState({ hasClock: false });
   };
 
-  leftClick = () => {
+  handleLeftClick = () => {
     this.setState({ hasClock: true });
   };
 
   render() {
+    const { hasClock, clockName } = this.state;
+
     return (
       <div className="App">
         <h1>React clock</h1>
-        {this.state.hasClock && (<Clock name={this.state.clockName} />)}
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,64 @@
...
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentDidUpdate(_prevProps: {}, prevState: State) {
+    if (prevState.clockName !== this.state.clockName && this.state.hasClock) {
+      // eslint-disable-next-line no-console
+      console.debug(`Renamed from ${prevState.clockName} to ${this.state.clockName}`);
+    }
+  }
```
vallerii, line 38: componentDidUpdate should be inside the Clock

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
-        </span>
+  leftClick = () => {
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {this.state.hasClock && (<Clock name={this.state.clockName} />)}
```
naziamoff, line 53: try state destructuring

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
-        {' time is '}
+  rightClick = (event: MouseEvent) => {
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  leftClick = () => {
```
naziamoff, line 45: follow the naming conventions from the checklist

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    return (
```
naziamoff, line 50: do you need return here?

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+  hasClock: boolean,
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
+
+  timerId = 0;
```
naziamoff, line 22: It is unclear what is stored inside the variable, think of a better name

src/App.tsx
naziamoff: actually in that case you will definitely need a return

naziamoff (changes requested):

src/App.tsx
```diff
@@ -19,38 +19,40 @@ export class App extends React.Component<{}, State> {
...
-    document.removeEventListener('click', this.leftClick);
+    document.removeEventListener('contextmenu', this.handleRightClick);
+    document.removeEventListener('click', this.handleLeftClick);
 
-    window.clearInterval(this.timerId);
+    window.clearInterval(this.nameChangeTimer);
   }
 
-  rightClick = (event: MouseEvent) => {
+  handleRightClick = (event: MouseEvent) => {
```
naziamoff, line 40: is much better but there's still a better way for naming your handlers)
I would come up with `hideClock/showClock`

src/App.tsx
```diff
@@ -19,38 +19,40 @@ export class App extends React.Component<{}, State> {
...
   };
 
   render() {
+    const { hasClock, clockName } = this.state;
+
     return (
       <div className="App">
         <h1>React clock</h1>
-        {this.state.hasClock && (<Clock name={this.state.clockName} />)}
+        {hasClock && (<Clock name={clockName} />)}
```
naziamoff, line 55: better to move the component to the new line

naziamoff (approved):
Well done, some minor comments though

PR #1035, opened 2022-10-19
Create a Clock class component

MAKSOti:
[DEMO LINK](https://MAKSOTi.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import { Clock } from './components/Clock';
+
 import './App.scss';
 
 function getRandomName(): string {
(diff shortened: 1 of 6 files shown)


AniaHna (commented):
Good job!

src/App.tsx
```diff
@@ -7,33 +9,67 @@ function getRandomName(): string {
...
+  clockName: string,
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  clockNameId = 0;
```
naziamoff, line 23: A name like that can easily confuse those who read. We don't store `nameId` in this variable
```suggestion
  clockNameTimerId = 0;
```

src/App.tsx
```diff
@@ -7,33 +9,67 @@ function getRandomName(): string {
...
-          {clockName}
-        </strong>
+  componentWillUnmount() {
+    window.clearInterval(this.clockNameId);
+  }
 
-        {' time is '}
+  addClock = () => {
+    this.setState({ hasClock: false });
+  };
```
naziamoff, line 54: the method `addClock` sets `hasClock` to false, but method `removeClock` sets it to true. That is counterintuitive, you should change that

src/components/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+      this.setState({ date: new Date() });
+    }, 1000);
+  }
+
+  componentDidUpdate(_: never, prevState: State) {
+    const { date } = this.state;
+
+    if (date !== prevState.date) {
+      window.console.info(date.toUTCString().slice(-12, -4));
+    }
```
naziamoff, line 29: there's a better way of doing this in order to make your code more clear.
My suggestion is to use [Date.toLocaleString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) method here

src/components/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {date.toUTCString().slice(-12, -4)}
```
naziamoff, line 49: You should follow DRY principle here. That means Don't Repeat Yourself. You already do it in your code, so why won't we create a separate method for that?

naziamoff (changes requested):

naziamoff (approved):
GJ! But you still repeat `date.toLocaleString().slice(-12, -4)`
Better to make it a separate variable

PR #1051, opened 2022-10-30
React clock

MeshackDev:
[DEMO LINK](https://MeshackDev.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


mayuces (approved):
Approved

src/App.tsx
```diff
@@ -1,39 +1,56 @@
...
+      this.setState({ hasClock: false });
+    });
+
+    document.addEventListener('click', () => {
+      this.setState({
+        hasClock: true,
+        clockName: getRandomName(),
+        today: new Date(),
+      });
+    });
```
ihor-jpeg, line 40: you should add event listeners in componentDidMount and remove them in component will unmount

src/Clock.tsx
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable no-console */
+import React from 'react';
+
+let timerId: number;
```
ihor-jpeg, line 4: should be inside of the class

ihor-jpeg (changes requested):

VinnieJ-2k20 (approved):

PR #1097, opened 2022-11-21
v1

DanilSerzh:
[DEMO LINK](https://DanilSerzh.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
+
+type State = {
+  hasClock: boolean,
+  clockName: string,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,62 @@
...
+  };
+
+  render() {
+    const { hasClock, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && <Clock clockName={clockName} />}
```
Esceype, line 58: If you use conditional rendering then it`s better to add a component at a new line after the logical operator and don`t forget to use brackets

Esceype (approved):
GJ ! Only one suggestion

src/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {today.toUTCString().slice(-12, -4)}
```
denys-danyliuk, line 49: you can move this to a variable

denys-danyliuk (approved):

src/App.tsx
denys-danyliuk: good point

PR #1100, opened 2022-11-21
v01

Anna-Shapovalova:
[DEMO LINK](https://Anna-Shapovalova.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,58 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string;
+};
+
+type State = {
+  today: Date;
+};
+
+export class Clock extends Component<Props, State> {
+  state: State = {
+    today: new Date(),
+  };
+
+  timerId = 0;
+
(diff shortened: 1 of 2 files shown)


IOneStarI (approved):
Nice!)

src/components/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const { today } = this.state;
+    const currentTime = today.toUTCString().slice(-12, -4);
+
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
```
Olexandr6, line 50:
```suggestion
        </strong>
        	time is
```

Olexandr6 (approved):

src/components/Clock.tsx
Anna-Shapovalova: I left it like this because of the spaces before and after 
Screenshot_57
Screenshot_56

PR #1104, opened 2022-11-22
Solution for clock

karyna-salnykova:
[DEMO LINK](https://karyna-salnykova.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,62 @@
+import { Component } from 'react';
+
+type Props = {
+  name: string,
+};
+
+type State = {
+  currentTime: Date,
+};
+
+export class Clock extends Component<Props, State> {
+  state: Readonly<State> = {
+    currentTime: new Date(),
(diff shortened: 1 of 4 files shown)


viktoriia-ovcharenko (commented):
something is wrong with the console. the changed name is displayed more times than it should. I think that's why you didn't pass one test

src/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+      this.setState({ currentTime });
+
+      // eslint-disable-next-line
+      console.info(this.getTimeFromDate(currentTime));
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Readonly<Props>) {
+    const { name } = this.props;
+
```
viktoriia-ovcharenko, line 31: you need to add a some condition here so that the name output in the console is displayed correctly

Olexandr6 (approved):
Well done 🔥

PR #1112, opened 2022-11-22
clock is ready

shtht24:
[DEMO LINK](https://shtht24.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,63 @@
+import { Component } from 'react';
+
+type State = {
+  today: Date,
+};
+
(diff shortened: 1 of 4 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {this.getTimeFromDate(today)}
```
denys-danyliuk, line 58: destructure that

PR #1125, opened 2022-11-30
Develop

SergIats:
[DEMO LINK](https://SergIats.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,65 @@
+import { Component } from 'react';
+
+function getRandomName(): string {
+  const value = Date.now().toString().slice(-4);
+
+  return `Clock-${value}`;
+}
+
+type State = {
+  clockName: string,
+  today: Date,
+};
+
+export class Clock extends Component<{}, State> {
+  state = {
+    clockName: getRandomName(),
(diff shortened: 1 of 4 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,65 @@
...
+      this.setState({ today: new Date() });
+      // eslint-disable-next-line
+      console.info(this.state.today.toUTCString().slice(-12, -4));
+    }, 1000);
+
+    window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+  }
+  // eslint-disable-next-line
```
vlad-demchuk, line 33: why do you need eslint disable here?

vlad-demchuk (changes requested):
check tests
image

src/Clock.tsx
SergIats: Hm... previously linter did't like empty object in componentDidUpdate. But without that object, I couldn't give previous clockName. I just thought that I can make this task without props, only with state. I hadn't can)) So, now I have added props and fixed it.

AlisaVlasova (approved):

PR #1170, opened 2022-12-29
Develop

halynarydasova:
[DEMO LINK](https://halynarydasova.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.scss';
+import { Clock } from './Components/clock/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
+type Props = {};
+type State = {
+  hasClock: boolean;
+  clockName: string;
+};
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+export class App extends React.Component<Props, State> {
```
polosanya, line 18:
```suggestion
export class App extends React.Component<{}, State> {
```
You can just make like this and don't create an empty type

src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
+    this.setState({ hasClock: false });
+  };
+
+  handleLeftClick = (event: MouseEvent) => {
+    event.preventDefault();
+    this.setState({ hasClock: true });
+  };
+
+  render(): React.ReactNode {
+    const { hasClock } = this.state;
```
polosanya, line 51:
```suggestion
    const { hasClock, clockName } = this.state;
```
Get all the props you need to make your code cleaner

src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
-        </span>
+export class App extends React.Component<Props, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  nameChangerId = 0;
+
+  componentDidMount(): void {
```
polosanya, line 26:
```suggestion
  componentDidMount() {
```
No need for that

polosanya (changes requested):
Almost perfect! Let's improve your code a bit)

polosanya (approved):
Well done!

PR #1175, opened 2022-12-30
Develop

vladimir-arkanov:
[DEMO LINK](https://vladimir-arkanov.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,8 @@
 import React from 'react';
 import './App.scss';
+import CircularProgress from '@mui/material/CircularProgress';
+import Box from '@mui/material/Box';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,39 +1,71 @@
...
-  }, 3300);
+export class App extends React.Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  counterRandomName = 0;
```
romanblyashuk94, line 24: name stores in state
this variable store id of started setTimeout
so name it "timerId"

src/App.tsx
```diff
@@ -1,39 +1,71 @@
...
-  );
-};
+  componentDidUpdate(): void {
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+  }
+
+  componentWillUnmount(): void {
+    document.addEventListener('contextmenu', (event) => {
```
romanblyashuk94, line 45: i think you wanted remove event listener, when component will unnmount.

also you should to pass to this method function that you passed to addEventListener.

now you pass two different functions.

put it to variable and pass variable to both methods

src/App.tsx
```diff
@@ -1,39 +1,71 @@
...
-        </span>
-      </div>
-    </div>
-  );
-};
+  componentDidUpdate(): void {
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+  }
```
romanblyashuk94, line 42: you create new listener for every update.
move it to componentDidMount and alsoo put callback function to variable

src/App.tsx
```diff
@@ -1,39 +1,71 @@
...
+          <h1 className="App__title">React clock</h1>
+        </div>
+
+        { hasClock && <Clock nameOfClock={clockName} /> }
+        { hasClock === false
+        && (
+          <Box className="Awaiting" sx={{ display: 'flex' }}>
+            <CircularProgress color="inherit" />
+          </Box>
+        ) }
```
romanblyashuk94, line 67: use ternarry operator here

hasClock ? clock : loader

romanblyashuk94 (changes requested):

src/App.tsx
vladimir-arkanov: Thanks for useful advices!!!

romanblyashuk94 (changes requested):
One more fix plz)

src/App.tsx
```diff
@@ -1,39 +1,67 @@
...
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+    return `Clock-${value}`;
+  };
 
-        {' time is '}
+  clickOrContextHandler = (event: MouseEvent) => {
```
romanblyashuk94, line 38: checklist p3
and I think that you should separate it to two diferent functions

```suggestion
  handleClickOrContext = (event: MouseEvent) => {
```
or better:

```suggestion
  handleClick = (event: MouseEvent) => {...}
  handleContext = (event: MouseEvent) => {...}
```

romanblyashuk94 (changes requested):
end check linter errors

src/App.tsx
valerii-derkach: fix these comments, please, rename - checklist p3 
and split logic

valerii-derkach (changes requested):
Good job! 

just fix method

ivan-kadykalo (approved):
Good job 🔥

PR #1178, opened 2022-12-30
Develop

TetKopot:
[DEMO LINK](https://TetKopot.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


romanblyashuk94 (changes requested):
Perfect solution. Just add some styles :)

romanblyashuk94 (approved):
👍

src/App.tsx
```diff
@@ -1,39 +1,81 @@
...
+import CssBaseline from '@mui/material/CssBaseline';
+import { Clock } from './Clock';
+
+const darkTheme = createTheme({
+  palette: {
+    mode: 'dark',
+  },
+});
+
+function App1() {
```
valerii-derkach, line 13: move it to a separate file and use another name, avoid using numbers in the name

valerii-derkach (changes requested):
look at console - your clock does not work correctly - it when clock is hiding the time continues running and if we click several times - hide/show the clock - console is running with old clock's time and the new time
also there are no clock names in console

valerii-derkach (approved):

PR #1187, opened 2023-01-02
Add solution

VladimirEdelmann:
# Task
Create a `Clock` class component that will update the time every second using a given markup:
- print current time on the page on page load;
  - use `.toUTCString().slice(-12, -4)` methods do awoid timezone issues;
- update the time every second using the `window.setInterval`;
- start the timer only when the component is added to the page (`componentDidMount`).- every second print the time in the DevTools using `console.info` method (**not** the `console.log`);
- make the `App` a class component;
- add the `hasClock` property to the `App` state;
- the `Clock` should be visible only when the `hasClock` is `true`;
- hide the `Clock` on a right mouse click in the `document` (`contextmenu` event):
    ```js
    document.addEventListener('contextmenu', (event) => {
      event.preventDefault(); // not to show the context menu

      // put your code here
    });
    ```
- show the `Clock` on a left mouse click in the `document` (`click` event):
    ```js
    document.addEventListener('click', () => {});
    ```
- the time should not be printed to the console when the Clock is hidden (`componentWillUnmount`);
- add the `clockName` having `Clock-0` default value to the `App` state;
- pass it to the `Clock` to be shown near the time (see the markup):
    ```jsx
    <Clock name={this.state.clockName} />
    ```
- update the `clockName` every `3300ms` with a new value generated by existing `getRandomName` function;
- each time the name is changed, the `Clock` must print a message with an old name and a new name using the `console.debug` method (use `componentDidUpdate`):
    ```
    Renamed from oldName to newName
    ```

# Implementation
[DEMO LINK](https://vladimiredelmann.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,69 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './components/Clock';
 import './App.scss';
 
-function getRandomName(): string {
+interface State {
+  clockName: string
+  hasClock: boolean
+  timerId: number
+}
+
+const getRandomName = (): string => {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
-}
-
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
-
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
-
-  // this code stops the timer
-  window.clearInterval(timerId);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
-
-        {' time is '}
-
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
-      </div>
-    </div>
-  );
 };
+
+export class App extends Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+    timerId: 0,
+  };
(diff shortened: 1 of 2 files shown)


kreisberg (approved):
Seems good!)

src/components/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface Props {
+  name: string
+}
+
+interface State {
+  currentTime: string
+}
+
+const toSlicedUTCDate = (date: Date = new Date()): string => {
```
solaryasha, line 11:
```suggestion
const convertToSlicedUTCDate = (date: Date = new Date()): string => {
```

src/App.tsx
```diff
@@ -1,39 +1,70 @@
...
+}
+
+const getRandomName = (): string => {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
-}
+};
+
+export class App extends Component<{}, State> {
```
solaryasha, line 16: [Let's add some fancy styles :)](https://mateacademy.slack.com/archives/C048KMNGNKU/p1672225786792459)

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,127 @@
...
+  };
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  getColorDigits = () => {
+    return this.state.clockName.slice(6);
+  };
+
+  handleColor = () => {
```
solaryasha, line 75: Good job 👍

solaryasha (approved):

PR #1201, opened 2023-01-03
solution

alyonkaradchuk:
[DEMO LINK](https://alyonkaradchuk.github.io/react_clock/)


APetrinko (changes requested):
All good, but you should add some styles for your clock

valerii-derkach (approved):
Good job!

PR #1214, opened 2023-01-04
react clock first write up

Kavster92:
[DEMO LINK](https://Kavster92.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
 import './App.scss';
+import { Component } from 'react';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


Iksich (approved):
Looks good, one thing you can just, import React from 'react';
Instead of Component

src/App.tsx
```diff
@@ -1,39 +1,64 @@
-import React from 'react';
 import './App.scss';
+import { Component } from 'react';
```
Iksich, line 2: import React from 'react';

Iksich (approved):

src/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type State = {
+  today: string;
+};
+
+type Props = {
+  clockName: string,
+};
+
+const formatedDate = (): string => {
```
naziamoff, line 11: function should be named as a function, starting with a verb that describes what the function does
```suggestion
const getFormattedDate = (): string => {
```

src/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return new Date().toUTCString().slice(-12, -4);
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    today: formatedDate(),
```
naziamoff, line 17: isn't this `now`?

naziamoff (approved):
Well done! Some minor comments

PR #1216, opened 2023-01-04
Solution

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { FC, Component } from 'react';
 import './App.scss';
 
 function getRandomName(): string {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.contextMenuHandler);
+    document.removeEventListener('click', this.clickHandler);
+  }
+
+  contextMenuHandler = (e: MouseEvent) => {
```
naziamoff, line 36: let's look through the checklist once again

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+
+  render() {
+    const { clockName, isClockVisible } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {
+          isClockVisible && <Clock clockName={clockName} />
+        }
```
naziamoff, line 53:
```suggestion
        {isClockVisible && (
          <Clock clockName={clockName} />
        )}
```

src/Components/Clock.tsx
```diff
@@ -0,0 +1,61 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string;
```
naziamoff, line 4: those are the Props of the `Clock` component, so we can name it just `name`

naziamoff (changes requested):

src/App.tsx
Rijo-J: changed method names from `clickHandler && contextMenuHandler` to `handleClick && handleContextMenu`

src/App.tsx
Rijo-J: ✅

src/Components/Clock.tsx
Rijo-J: ✅

naziamoff (approved):
🔥

PR #1232, opened 2023-01-17
final

Alireza-Mohammadhossein:
[DEMO LINK](https://Alireza-Mohammadhossein.github.io/react_clock/)


meforoma (approved):
it'll do
in the hook implementation don't forget to
add the `console.info` upon clock rename

PR #1241, opened 2023-01-23
React Clock - completed

GarryRocksman:
[DEMO](https://GarryRocksman.github.io/react_clock/)

 - Have a nice Day 🔥

### DIFF
--- a/src/Components/Clock.tsx
+++ b/src/Components/Clock.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+
+type Props = {
+  name: string,
+};
+export class Clock extends React.Component<Props> {
+  state = {
+    today: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({
+        today: new Date(),
+      });
+      // eslint-disable-next-line no-console
+      console.info(this.state.today.toUTCString()
+        .slice(-12, -4));
+    }, 1000);
+  }
+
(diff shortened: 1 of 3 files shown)


src/Components/Clock.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+
+type Props = {
+  name: string,
+};
+export class Clock extends React.Component<Props> {
+  state = {
+    today: new Date(),
+  };
```
nzrii-kysel, line 9: create type for state

src/Components/Clock.tsx
```diff
@@ -0,0 +1,53 @@
...
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({
+        today: new Date(),
+      });
+      // eslint-disable-next-line no-console
+      console.info(this.state.today.toUTCString()
+        .slice(-12, -4));
```
nzrii-kysel, line 20: you do this formating twice, I'd create a separate function for it

nzrii-kysel (changes requested):

src/Components/Clock.tsx
GarryRocksman: ep, fixed

src/Components/Clock.tsx
GarryRocksman: Good point,  added normalize date function,

GarryRocksman:
fixed

alexandra-protyanova (approved):
🚀

PR #1247, opened 2023-01-24
Clock was added

svitjojo:
[DEMO LINK](https://svitjojo.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
+
+type State = {
+  hasClock: boolean;
+  clockName: string;
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,56 @@
...
+type State = {
+  today: Date,
+};
+
+type Props = {
+  name: string,
+};
+
+export class Clock extends Component<Props, State> {
+  state: Readonly<State> = {
```
OlhaYevstifieieva, line 12: I'm not sure we need extra type of 'state', but I could be wrong

OlhaYevstifieieva (approved):
Looks nice!

src/components/Clock/Clock.tsx
alexandra-protyanova: you are changing state on 22 line, I think it is not readonly

alexandra-protyanova, line 12:
```suggestion
  state: State = {
```

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
```
alexandra-protyanova, line 17:
```suggestion
  state: State = {
```

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 💫

PR #1254, opened 2023-01-26
add solution

Anastasiia-Khudych:
[Demo](https://anastasiia-khudych.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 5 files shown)


victorionov (approved):
super

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
+  clockName: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  setName = 0;
```
Anna-Tkachenko, line 22: it's not clear what exactly stored in this variable
will be better to add `timer` to the variable name

src/components/clock/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+  timerId = 0;
+
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({
+        date: new Date(),
+      });
+
+      // eslint-disable-next-line no-console
+      console.info(this.state.date.toUTCString().slice(-12, -4));
```
Anna-Tkachenko, line 25: will be better to create separate function for getting date in needed format, because you use the same functionality twice in your code

Anna-Tkachenko (approved):

PR #1259, opened 2023-01-27
 react_clock

rafmamedov:
[DEMO LINK](https://rafmamedov.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -50,8 +50,10 @@ export class Clock extends Component<Props, State> {
           {clockName}
         </strong>
 
+        {' time is '}
+
         <span className="Clock__time">
-        { ` time is ${today}` }
+          {today}
         </span>
       </div>
     );


src/App.tsx
```diff
@@ -1,39 +1,62 @@
...
+  };
+
+  render() {
+    const { hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && (
+          <Clock clockName={this.state.clockName} />)}
```
loralevitska, line 58:
```suggestion
        {hasClock && (
          <Clock clockName={this.state.clockName} />
        )}
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {today}
+        </span>
```
loralevitska, line 57: What is this span for if the class is not used?
```suggestion
        {`  time is  ${today}`}
```

loralevitska (changes requested):
Almost done)
Let's improve

src/components/Clock.tsx
rafmamedov: This span with className="Clock__time" is needed in tests. Without this, tests fail.
Знімок екрана 2023-01-28 о 11 36 21

src/App.tsx
rafmamedov: Thanks!

loralevitska (approved):

PR #1264, opened 2023-01-28
Solution

maasiimka:


svitjojo (changes requested):
Create a components folder, there is a Clock folder and put Clock.tsx there. Also, create index.ts and export Clock inside.

src/App.tsx
```diff
@@ -1,39 +1,73 @@
...
+    this.setState({ hasClock: false });
+  };
+
+  render(): ReactNode {
+    const { hasClock, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && <Clock clockName={clockName} />}
```
Anna-Tkachenko, line 69:
```suggestion
        {hasClock && (
          <Clock clockName={clockName} />
        )}
```

src/Components/Clock/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  clockId: number | null = null;
+
+  componentDidMount() {
+    this.clockId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      // eslint-disable-next-line no-console
+      console.info(this.state.today.toUTCString().slice(-12, -4));
```
Anna-Tkachenko, line 22: will be better to create separate function for transforming date, because for now you repeated the same code twice

Anna-Tkachenko (approved):

PR #1265, opened 2023-01-28
Solution

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
+
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
@@ -7,33 +9,52 @@ function getRandomName(): string {
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+type State = {
+  hasClock: boolean;
+  clockName: string,
+};
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
+
+  timerId = window.setInterval(() => {
+    this.setState({ clockName: getRandomName() });
   }, 3300);
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentDidMount() {
+    document.addEventListener('contextmenu', (ev) => {
+      ev.preventDefault();
+      this.hideClock();
+    });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => {
+      this.appearClock();
+    });
+  }
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  componentWillUnmpunt() {
+    clearInterval(this.timerId);
+  }
 
-        {' time is '}
+  hideClock() {
+    this.setState({ hasClock: false });
+  }
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  appearClock() {
+    this.setState({ hasClock: true });
+  }
+
+  render() {
+    const { hasClock, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && <Clock name={clockName} />}
       </div>
-    </div>
-  );
-};
+    );
+  }
+}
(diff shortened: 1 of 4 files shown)


qoa11a (changes requested):
let's work on passing all the tests
image

src/App.scss
```diff
@@ -1,8 +1,49 @@
+/* eslint-disable max-len */
+
 iframe {
   display: none;
 }
```
Volodymyr-Melnyk-Fron10der, line 5: 1. use class selectors instead of tag selectors (you could only if there is no other way)
2. I don't see you use iframe in you JSX, why do you need to style it?

src/App.scss
```diff
@@ -1,8 +1,49 @@
+/* eslint-disable max-len */
+
```
Volodymyr-Melnyk-Fron10der, line 2: Fix eslint errors, not just ignore them)

src/App.scss
```diff
@@ -1,8 +1,49 @@
...
+  rgba(223,233,243,1) 0%, rgba(255,255,255,1) 100%);
+  --font-main: 24px;
+  padding-top: 20vh;
+
+  font-family: "Roboto", sans-serif;
+  font-size: var(--font-main);
+  letter-spacing: 3px;
+  color: #94b0da;
+  background: var(--bg-main);
+}
```
Volodymyr-Melnyk-Fron10der, line 24: 1. use class selector
2. I recommend do not style body as much, it is better to create new inside container as `<div class="wrapper">`

src/App.scss
```diff
@@ -1,8 +1,49 @@
...
+
+.App {
```
Volodymyr-Melnyk-Fron10der, line 26: App is only one selector which is belong to the file `App.scss`,  other selectors should be in separated files like reset.scss, mixins.scss...

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+type State = {
+  hasClock: boolean;
+  clockName: string,
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
```
Volodymyr-Melnyk-Fron10der, line 21: initialize state inside constructor

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
+  timerId = window.setInterval(() => {
+    this.setState({ clockName: getRandomName() });
   }, 3300);
```
Volodymyr-Melnyk-Fron10der, line 25: 1. I would rather move timerId  into state
2. Start interval inside componentDidMount

src/components/Clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+type Props = {
+  name: string,
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    today: new Date(),
+  };
+
```
Volodymyr-Melnyk-Fron10der, line 16: initialize state inside constructor

src/components/Clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+  timerId = 0;
```
Volodymyr-Melnyk-Fron10der, line 17: move into state

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, Great job! I left few comments

src/App.scss
IVxVI: fixed

src/App.scss
IVxVI: fixed)

src/App.scss
IVxVI: fixed)

src/App.scss
IVxVI: fixed)

src/App.tsx
IVxVI: Had so many troubles with it. I couldn`t assign to timerId further, cuz it`s ReadOnly? also some type errors and eslint asks for usage as class property, not inside constructor. Is it really bertter option? In this video https://mate.academy/learn/react/class-components#/video/470 Misha says that it is fine to write as I did?

src/App.tsx
IVxVI: fixed

src/App.tsx
```diff
@@ -1,39 +1,64 @@
...
+    this.state.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmpunt() {
+    window.clearInterval(this.state.timerId);
```
loralevitska, line 41: Better clear listeners here too.

loralevitska (changes requested):
You did amazing job. Almost perfect) Let's improve

src/App.tsx
IVxVI: TY, fixed)

SanyaBratashchuk (approved):

PR #1269, opened 2023-01-29
solution

metalikaqq:


src/App.tsx
```diff
@@ -1,39 +1,58 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock/Clock';
```
svitjojo, line 3:
```suggestion
import { Clock } from './Clock';
```

svitjojo (changes requested):
The code is good 👍 
Your demo link is not working, check it.
In Clock folder create index.ts and export Clock.tsx there. After inside App.tsx you can import without double name repeating

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+// eslint-disable-next-line react/prefer-stateless-function
+export class App extends Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
```
Anna-Tkachenko, line 25: you need to remove all handlers and reset timers in `componentWillUnmount` method

Anna-Tkachenko: you didn't remove event listeners

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+  handleClick = () => {
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {this.state.hasClock && <Clock name={this.state.clockName} />}
```
Anna-Tkachenko, line 54: will be better to get needed properties from state through desctructuring before the `return` statement

Anna-Tkachenko (changes requested):

PR #1271, opened 2023-01-29
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,70 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './components/Clock';
+
 import './App.scss';
 
-function getRandomName(): string {
-  const value = Date.now().toString().slice(-4);
-
-  return `Clock-${value}`;
+interface State {
+  clockName: string;
+  id: number;
+  hasClock: boolean;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    id: 0,
+    hasClock: true,
+  };
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+  componentDidMount() {
+    const timerId = window.setInterval(() => {
+      this.getRandomName();
+    }, 3300);
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+    this.setState({ id: timerId });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+  }
 
-        {' time is '}
+  componentWillUnmount() {
+    window.clearInterval(this.state.id);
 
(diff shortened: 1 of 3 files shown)


NVyrsta (approved):
Good job 👍

src/App.tsx
```diff
@@ -1,39 +1,70 @@
...
+  hasClock: boolean;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    id: 0,
```
Anna-Tkachenko, line 15: if the variable doesn't affect on the content of the component, you can create it directly in the class, not in the state object

src/App.tsx
```diff
@@ -1,39 +1,70 @@
...
-    clockName = getRandomName();
-  }, 3300);
+  componentWillUnmount() {
+    window.clearInterval(this.state.id);
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+    document.removeEventListener('contextmenu', () => {
+      this.setState({ hasClock: false });
+    });
```
Anna-Tkachenko, line 41: function won't be removed in this way
for deleting handlers you need to pass the same link to the same function as the second argument
for now you created separate function, and pass it here, it won't delete existing handler

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+  hideClock: EventListener = (event) => {
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
+
+  showClock: EventListener = () => {
+    this.setState({ hasClock: true });
+  };
+
+  changeVisability() {
```
Anna-Tkachenko, line 52: do you need this function? looks like you don't use it

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+  }
+
+  render() {
+    const { clockName, hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && <Clock name={clockName} />}
```
Anna-Tkachenko, line 63:
```suggestion
        {hasClock && (
          <Clock name={clockName} />
        )}
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface State {
+  today: Date;
+  id: number;
+}
+
+export class Clock extends Component<Props, State> {
+  state = {
+    today: new Date(),
+    id: 0,
```
Anna-Tkachenko, line 15: what's the purpose of this property in state?

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+
+  render() {
+    const { clockName, hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && (
+
```
Anna-Tkachenko, line 60: empty lines are redundant before and after `Clock`

src/components/Clock.tsx
Anna-Tkachenko: you didn't delete this property
what's the reason of adding `id` in state?

Anna-Tkachenko (changes requested):

src/components/Clock.tsx
imariash86: there is no reason, sorry, I just didn't notice this comment before

Anna-Tkachenko (approved):
Good job!

PR #1288, opened 2023-01-30
 solution

OlehSharha:
[DEMO LINK](https://OlehSharha.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/clock';
 
 function getRandomName(): string {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+    });
+  };
+
+  render() {
+    const { hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && <Clock name={this.state.clockName} />}
```
Anna-Tkachenko, line 64: will be better to desctucture needed property from state before the `return` statement

Anna-Tkachenko (approved):
⭐️

PR #1292, opened 2023-01-30
Solution

alSemenyuk:
https://alSemenyuk.github.io/react_clock/

### DIFF
--- a/src/component/Clock.tsx
+++ b/src/component/Clock.tsx
@@ -1,0 +1,54 @@
+import { Component } from 'react';
+
+type Props = {
+  name: string;
+};
+
+type State = {
+  date: Date;
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    date: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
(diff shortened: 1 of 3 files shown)


PolinaSukhorukova (changes requested):
GJ 🔥 
you have to do some fixes

src/component/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  render() {
+    const { date } = this.state;
+
+    return (
+      <div className="Clock">
+
+        <strong className="Clock__name">
+          {this.props.name}
```
Anna-Tkachenko, line 44: will be better to get needed properties from the `props` object via destructuring before the `return` statement

src/component/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+    date: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+      // eslint-disable-next-line no-console
+      console.info(this.state.date.toUTCString().slice(-12, -4));
```
Anna-Tkachenko, line 22: will be better to create separate function for transforming date, for now you used the same code twice

Anna-Tkachenko (approved):
Good job!

PR #1293, opened 2023-01-30
Solution by DK

DimaK123:


polosanya (changes requested):
Please add the Clock component to your commit. Now it's absent on the GitHub
image

src/App.tsx
```diff
@@ -7,33 +8,63 @@ function getRandomName(): string {
...
+  componentDidUpdate(_prevProps: {}, prevState: State) {
+    const { hasClock, clockName } = this.state;
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    if (hasClock === true
+      && clockName !== prevState.clockName) {
+      // eslint-disable-next-line no-console
+      console.info(`Renamed from ${prevState.clockName} to ${clockName}`);
```
qoa11a, line 40: it should be console.debug

qoa11a (changes requested):

src/App.tsx
DimaK123: fixed

agniya-i (approved):

PR #1313, opened 2023-02-13
Solves

Vasyliy22:
[DEMO LINK](https://Vasyliy22.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+    event.preventDefault();
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+    this.setState({ hasClock: false });
+  };
+
+  render() {
+    const { hasClock } = this.state;
```
witflash, line 49: If you use destructuring, you can grab the `clockName` constant here and use it below in JSX.

witflash:
```
const { clockName, hasClock } = this.state;
```

src/Clock.tsx
```diff
@@ -0,0 +1,56 @@
...
+  today: Date,
+};
+
+type Props = {
+  name: string,
+};
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    today: new Date(),
```
witflash, line 13: It should be better the `currentTime` name.

witflash (approved):
Good job 👍🏼
Check my comments to improve your future work.

PR #1314, opened 2023-02-14
React_clock

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,70 @@
+import React from 'react';
+
+function getRandomName(): string {
+  const value = Date.now().toString().slice(-4);
+
+  return `Clock-${value}`;
+}
+
+function stringDate(info: Date) {
+  return info.toUTCString().slice(-12, -4);
+}
+
+type State = {
+  today: Date;
+  clockName: string;
+};
+
+type Props = {
+  status: boolean;
+};
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    today: new Date(),
+    clockName: 'Clock-0',
+  };
+
+  componentDidMount(): void {
+    window.setInterval(this.timerId, 3300);
+    window.setInterval(this.updateDate, 1000);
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,70 @@
...
+      console.info(stringDate(this.state.today));
+    }
+  };
+
+  render() {
+    return (
+      <>
+        {this.props.status && (
+          <div className="Clock">
+            <strong className="Clock__name">{this.state.clockName}</strong>
```
OleziO, line 60: Don't use upper case for the first letter, better practice is to start the name from the lower letter. Check other similar cases.

OleziO (changes requested):

src/components/Clock.tsx
```diff
@@ -0,0 +1,70 @@
...
+    window.setInterval(this.updateDate, 1000);
+  }
+
+  componentDidUpdate(
+    _prevProps: Readonly<{}>,
+    prevState: Readonly<{ clockName: string }>,
+  ): void {
+    if (prevState.clockName !== this.state.clockName) {
+      // eslint-disable-next-line no-console
+      console.debug(`Renamed from ${prevState.clockName} to ${this.state.clockName} `);
```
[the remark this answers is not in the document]
07Spirit: This test fall if I delete this message. But with it, I haven't problem with linter.

src/components/Clock.tsx
07Spirit: This names was created in empty project. If I change this names, use lower case, all test fall.

polosanya (changes requested):
The clock name should be updated only when the clock is visible
image

07Spirit:
In test write next: "should print new name with console.debug each time it is updated after it was hidden".
I do this.

qoa11a (approved):

PR #1316, opened 2023-02-14
solution

Deniskakaka:
https://deniskakaka.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,56 @@
...
+      this.setState({ showClock: false });
+    });
+
+    document.addEventListener('click', () => {
+      this.setState({ showClock: true });
+    });
+  }
+
+  componentDidUpdate(prevProps:State) {
+    console.log(
```
mykhalenych, line 39: you have lint error here, change to debug

src/App.tsx
```diff
@@ -1,39 +1,56 @@
...
+
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {this.state.showClock && <Clock name={this.state.clockName} />}
```
mykhalenych, line 52: destruct your state

mykhalenych (changes requested):

src/App.tsx
Deniskakaka: i added eslint-disable-next-line no-console, if it is correct, if no what exelent i should add?

mykhalenych: console.debug

mykhalenych (changes requested):
delete eslint and add console.debug

etojeDenys (approved):
LGTM

PR #1355, opened 2023-03-10
Solution

Dynamizx:
https://dynamizx.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,8 +20,8 @@ export class App extends Component<{}, State> {
   timerID = 0;
 
   componentDidMount() {
-    document.addEventListener('click', this.handleDocumentClick);
-    document.addEventListener('contextmenu', this.handleDocumentContextmenu);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,39 +1,56 @@
...
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    const { clockName, hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && (<Clock name={clockName} />)}
```
witflash, line 52: FYI
You can use this syntax without additional brackets if you have single-line conditional render:
```
```
It's not an error but it looks better :)

witflash (changes requested):
Hey!
Your code looks good 👍 
But can you check if you successfully deploy your project? Because your DEMO LINK doesn't work :(

Anastasiia-Svintsova (approved):

PR #1357, opened 2023-03-11
Solution

Vlad-Fedorishchev:
[DEMO LINK](https://Vlad-Fedorishchev.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,54 @@
+import { Component } from 'react';
+import './App.scss';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,54 @@
+import { Component } from 'react';
+import './App.scss';
+
+type Props = {
```
agniya-i, line 4: it would be nice to have type for State as well

agniya-i (approved):

PR #1371, opened 2023-03-16
Solution

VadymTsyndra:
https://VadymTsyndra.github.io/react_clock/


IvanFesenko (changes requested):
Time should not diplicate
2023-03-16_16h17_32

PR #1376, opened 2023-03-16
React clock

Anton-Tretiak:
[DEMO LINK](https://anton-tretiak.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,65 @@
...
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+type State = {
+  clockName: string;
+  displayClock: boolean;
```
Lilu580, line 13: its should be named like hasClock its was in README file,but I think its also good name

src/App.tsx
```diff
@@ -1,39 +1,65 @@
...
+    this.setState({ clockName: getRandomName() });
+  };
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  handleLeftClick = () => {
+    this.setState({ displayClock: true });
+  };
```
Lilu580, line 44: Im not sure but im think we should use here event.preventDefault();

Lilu580 (approved):
Good solution

PR #1388, opened 2023-03-16
react-clock done

Stepan-TS:
[DEMO LINK](https://Stepan-TS.github.io/react_clock/)

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -1,0 +1,58 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string,
+};
+
+type State = {
+  date: Date,
+};
+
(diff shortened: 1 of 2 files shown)


src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+    date: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+
+      window.console.info(dateFormat(this.state.date));
```
OleksiiBlezniuk, line 24: Don't use state here, it can be outdated
Save `new Date()` result to constant and reuse it for both setState and console.info

OleksiiBlezniuk (approved):

PR #1389, opened 2023-03-16
Develop

Ambulence:
-[DEMO LINK](https://Ambulence.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+
+interface Props {
+  name: string;
+}
+
+type State = {
+  today: Date;
+};
+
+export class Clock extends React.Component<Props, State> {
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+  state: Readonly<State> = {
+    today: new Date(),
+  };
+
+  timer = 0;
+
+  componentDidMount() {
+    this.timer = window.setInterval(() => {
+      this.setState({ today: new Date() });
+      window.console.info(correctTime(this.state.today));
```
OleksiiBlezniuk, line 25: don't use state here, it can be outdated
save `new Date` result to constant and use it in both setState and console.info

OleksiiBlezniuk (approved):
Will approve this time, but pay attention to comment

PR #1390, opened 2023-03-16
Solution

sasha-dranuta:
[DEMO LINK](https://sasha-dranuta.github.io/react_clock/)

### DIFF
--- a/src/Components/Clock.tsx
+++ b/src/Components/Clock.tsx
@@ -1,0 +1,58 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string;
+};
+
+type State = {
+  date: Date,
+};
+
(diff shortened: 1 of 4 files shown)


src/Components/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  clockId = 0;
+
+  componentDidMount() {
+    this.clockId = window.setInterval(() => {
+      this.setState({ date: new Date() });
+
+      window.console.info(
+        this.state.date.toUTCString().slice(-12, -4),
+      );
```
OleksiiBlezniuk, line 24: Don't use state here, because it can be outdated, save result of `new Date()` to constant and use it here

OleksiiBlezniuk (approved):

PR #1397, opened 2023-03-17
Solution

DenysDubas:
(https://DenysDubas.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,62 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string;
+};
+
+type State = {
+  today: Date;
+};
+
+const getNormTime = (date: Date) => (
+  date.toUTCString().slice(-12, -4)
+);
+
+export class Clock extends Component<Props, State> {
+  state = {
+    today: new Date(),
+  };
+
+  clockId = 0;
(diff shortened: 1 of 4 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+  clockName: string;
+};
+
+type State = {
+  today: Date;
+};
+
+const getNormTime = (date: Date) => (
+  date.toUTCString().slice(-12, -4)
+);
```
OleksiiBlezniuk, line 13: norm

OleksiiBlezniuk: I guess this method should have other name =)

src/components/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  clockId = 0;
+
+  componentDidMount() {
+    this.clockId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      window.console.info(
+        getNormTime(this.state.today),
+      );
```
OleksiiBlezniuk, line 28: Don't use this state value here, it can be outdated
Save result of `new Date()` to constant and use it in this function

OleksiiBlezniuk (approved):

PR #1400, opened 2023-03-18
solution

IhorChaikovskyi:
[DEMO LINK](https://IhorChaikovskyi.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,70 @@
...
+};
+
+export class App extends Component<{}, State> {
+  // // This code starts a timer
+  // const timerId = window.setInterval(() => {
+  //   clockName = getRandomName();
+  // }, 3300);
+
+  // // this code stops the timer
+  // window.clearInterval(timerId);
```
OleziO, line 24: Remove all comments from the code.

OleziO (approved):

src/App.tsx
IhorChaikovskyi: Ok.I've been doing this code for so long that I forgot about these comets)

PR #1402, opened 2023-03-19
Develop

kuzmi491:
https://kuzmi491.github.io/react_clock

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,67 @@
-import React from 'react';
+// import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
-  const value = Date.now().toString().slice(-4);
+  const value = Date.now()
+    .toString()
+    .slice(-4);
 
   return `Clock-${value}`;
 }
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,67 @@
...
-          {clockName}
-        </strong>
+  rightClick = (event: Event) => {
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
 
-        {' time is '}
+  leftClick = (event: Event) => {
+    event.preventDefault();
```
OleksiiBlezniuk, line 48: I think this preventing not needed here

src/Clock.tsx
```diff
@@ -0,0 +1,69 @@
...
+    today: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      window.console.info(fixTime(this.state.today));
```
OleksiiBlezniuk, line 26: Don't use state here, it can be outdated
Save `new Date` result to constant and use it in both setState and console.info

OleksiiBlezniuk (approved):

PR #1403, opened 2023-03-19
Solution clock

nazarovoleksiy:
[Demo](https://nazarovoleksiy.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,62 @@
+import { Component } from 'react';
+
+type State = {
+  today: Date;
+};
+
+type Props = {
+  clockName: string;
+};
+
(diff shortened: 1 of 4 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  clockId = 0;
+
+  componentDidMount() {
+    this.clockId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      window.console.info(
+        getFormattedTime(this.state.today),
+      );
```
OleksiiBlezniuk, line 28: Don't use state here, it can be outdated
Save new Date result to constant and use it for both setState and console.info

OleksiiBlezniuk (approved):

PR #1407, opened 2023-03-19
added solution

vladyslav-penkin:
- [DEMO LINK](https://Vlados3611.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,62 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string;
+};
+
+type State = {
+  today: Date;
+};
+
+function getTime(date: Date): string {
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  clockId = 0;
+
+  componentDidMount(): void {
+    this.clockId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      window.console.info(
+        getTime(this.state.today),
+      );
```
OleksiiBlezniuk, line 28: Don't use state here, it can be outdated
Save `new Date()` result to constant and use if for both `setState` and `console.info`

OleksiiBlezniuk (approved):

PR #1412, opened 2023-03-22
React clock

VadKol:
- [DEMO LINK](https://VadKol.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  render() {
+    const { hasClock, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && <Clock clockName={clockName} />}
```
sergii-nosachenko, line 55:
```suggestion

        {hasClock && (
          <Clock clockName={clockName} />
        )}
```

src/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+  state: Readonly<State> = {
+    today: new Date(),
+  };
+
+  timerCountId = 0;
+
+  componentDidMount() {
+    this.timerCountId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+      window.console.info(formatTime(this.state.today));
```
sergii-nosachenko, line 25: It will print the previous value, not the current one. Create a constant for the value and pass it to setState and console.

https://legacy.reactjs.org/docs/react-component.html#setstate

sergii-nosachenko (changes requested):

src/App.tsx
VadKol: Fixed ✔

src/Clock.tsx
VadKol: Thank you) Lost it)
Fixed ✔

src/Clock.tsx
```diff
@@ -21,8 +21,14 @@ export class Clock extends Component<Props, State> {
...
-      this.setState({ today: new Date() });
-      window.console.info(formatTime(this.state.today));
+      const today = new Date();
+
+      this.setState(
+        { today },
+        () => {
+          window.console.info(formatTime(this.state.today));
+        },
+      );
```
sergii-nosachenko, line 31: Cool solution 🔥

You can even omit to create a variable cause the callback function will get an updated value.
```suggestion
      this.setState(
        { today: new Date() },
        () => {
          window.console.info(formatTime(this.state.today));
        },
      );
```

sergii-nosachenko (approved):
Good work 👍

PR #1415, opened 2023-03-23
ReactClock

Anatolii-Polishchuk:
https://Anatolii-Polishchuk.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,56 @@
 import React from 'react';
 import './App.scss';
 
+import { Clock } from './Components/Clock';
+
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+type State = {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,56 @@
 import React from 'react';
 import './App.scss';
 
+import { Clock } from './Components/Clock';
```
IvanFesenko, line 4: Where is that file?

IvanFesenko (changes requested):
Please push all needed files to github

src/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {this.state.today.toUTCString().slice(-12, -4)}
```
polosanya, line 49: Would be great to make it a variable in order not to duplicate yourself, as you use it at least twice in your code

polosanya (changes requested):
Almost done!

IvanFesenko (approved):

PR #1420, opened 2023-03-28
commit message

diamat1997:
[DEMO LINK](https://diamat1997.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,51 @@
...
+      action.preventDefault();
+
+      this.state.hasClock = false;
+    });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => {
+      this.state.hasClock = true;
```
anstsot, line 31: please change the state by using `this.setState` because it's not working on demo

anstsot (changes requested):
Good work. Please fix the issue with state updating

PR #1422, opened 2023-03-29
react_clock

MaksymHuriev:
[DEMO LINK](https://MaksymHuriev.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,56 @@
+/* eslint-disable no-console */
+import { Component } from 'react';
+
+export interface Props {
+  name: string;
+}
+
+export interface State {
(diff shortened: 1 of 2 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable no-console */
+import { Component } from 'react';
+
+export interface Props {
+  name: string;
+}
+
+export interface State {
```
SanyaBratashchuk, line 8:
```suggestion
interface Props {
  name: string;
}

interface State {
```
Why do you need to export this interfaces?

SanyaBratashchuk (approved):

PR #1426, opened 2023-04-03
solution :)

katerynashylina:
[DEMO LINK](https://katerynashylina.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { CLock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,58 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { CLock } from './Clock';
```
ivbutsykin, line 3: Hm, I don't see this file in your commit. Did you add it?

Plus, make attention naming - should be `<Clock />`

ivbutsykin (changes requested):

src/App.tsx
katerynashylina: I again forgot to, I am so sorry, let me fix it. And I didn't really get what naming you are talking about..

Volodymyr-Melnyk-Fron10der (changes requested):

src/App.tsx
katerynashylina: Oh yeah, I see now, so careless..

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+  }
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+  componentWillUnmount() {
+    document.removeEventListener('click', this.handleLeftClick);
+    document.removeEventListener('contextmenu', this.handleRigthClick);
+  }
```
polosanya, line 29: Don't forget to remove the interval too
image

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  clockNameChanger = () => {
```
polosanya, line 31:
```suggestion
  changeClockName = () => {
```
image

src/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type State = {
+  currentTime: Date,
+};
+
+type Props = {
+  clockName: string;
+};
+
+const avoidingTimezoneIssues = (time: Date) => {
```
polosanya, line 11: I think something like `formatTime` will sound better

polosanya (changes requested):
Almost perfect!

SerhiiKirik (approved):
Great 🔥

PR #1435, opened 2023-04-07
solution

MSA1991:
[DEMO LINK](https://msa1991.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,13 @@
-import React from 'react';
+import { Component } from 'react';
+
+import { Clock } from './components/Clock';
+
 import './App.scss';
+
+type State = {
+  hasClock: boolean,
+  clockName: string,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,71 @@
...
-        </span>
+      this.setState({ hasClock: false });
+    });
+
+    document.removeEventListener('click', (event) => {
+      event.preventDefault();
+
+      this.setState({ hasClock: true });
+    });
+  }
```
Volodymyr-Melnyk-Fron10der, line 58: I wouldn't work)
You have to pass a link to the function, so it will be the same and method removeEventListener will understand wich function you want to remove. BUT now you have passed one function in addEventListener and different one in removeEventListener

image

Volodymyr-Melnyk-Fron10der: create separated methods in the class and use them

src/components/Clock.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
+  }
+
+  render() {
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.name}
```
Volodymyr-Melnyk-Fron10der, line 49: use desctruture assignment 

const { name } = this.props

Volodymyr-Melnyk-Fron10der (changes requested):
almost there)

Anastasiia-Svintsova (approved):

PR #1449, opened 2023-04-12
Solution

RomaSheva1987:
[DEMO LINK](https://RomaSheva1987.github.io/react_clock/)

### DIFF
--- a/src/Clock/Clock.tsx
+++ b/src/Clock/Clock.tsx
@@ -1,0 +1,57 @@
+import { Component } from 'react';
+
+type State = {
+  today: Date,
+};
+
+type Props = {
+  name: string,
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    today: new Date(),
+  };
+
+  timerId = 0;
+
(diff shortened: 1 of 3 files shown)


src/Clock/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+    this.timerId = window.setInterval(() => {
+      const today = new Date();
+
+      this.setState({ today });
+      // eslint-disable-next-line no-console
+      console.info(today.toUTCString().slice(-12, -4));
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Props) {
```
etojeDenys, line 28: Instead of comparing entire `prevProps` and `this.props`, you should compare individual prop values (in this case, `prevProps.name !== this.props.name`) to avoid unnecessary console logs when other props change in the future. 🧐

etojeDenys (changes requested):
almost done

sTorba24 (approved):
Well done!

PR #1450, opened 2023-04-12
Solution react_clock

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,59 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,59 @@
-import React from 'react';
+import { Component } from 'react';
```
etojeDenys, line 1: Consider importing React as well since you're using JSX in your render method. For example: import React, { Component } from 'react';

etojeDenys (approved):
lgtm

PR #1459, opened 2023-04-15
Solution

Anna-Kuzmenko:
[DEMO LINK](https://Kuzmenko-Anna.github.io/react_clock/)

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+
+type Props = {
+  clockName: string;
+};
+
+interface State {
+  today: string;
+}
+
+export class Clock extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    today: new Date().toUTCString().slice(-12, -4),
+  };
+
+  timerId: number | null = null;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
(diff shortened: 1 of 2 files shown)


vinnykalisa (approved):
Good job! Nice solution!
You can also extract just the Component from React because linter on git doesn't complain as linter in VS Code does, so you won't see this error: 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+  clockName: string;
+};
+
+interface State {
+  today: string;
+}
+
+export class Clock extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    today: new Date().toUTCString().slice(-12, -4),
```
SerhiiKirik, line 13: DRY
you do this twice, let's move this to the separate function

SerhiiKirik (changes requested):
Great 🔥 
Let's deploy your solution & improve your solution 🧑‍💻

src/components/Clock/Clock.tsx
Anna-Kuzmenko: Moved :white_check_mark:

SerhiiKirik (approved):
LGTM 🔥

PR #1467, opened 2023-04-17
develop

glots-a:
[DEMO LINK](https://glots-a.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Clock } from './components/Clock';
 import './App.scss';
 
 function getRandomName(): string {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -7,33 +8,49 @@ function getRandomName(): string {
...
-          {clockName}
-        </strong>
+export class App extends React.Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
 
-        {' time is '}
+  timerId = window.setInterval(() => {
```
yuliia-anisimova, line 22: I think it's better to set the interval in componentDidMount() instead of directly defining timerId as a class property.

yuliia-anisimova (approved):
Great job!

src/App.tsx
```diff
@@ -7,33 +8,49 @@ function getRandomName(): string {
...
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  componentDidMount() {
+    document.addEventListener('contextmenu', this.handleRightClick);
+    document.addEventListener('click', this.handleLeftClick);
+  }
+
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+    document.addEventListener('contextmenu', this.handleRightClick);
```
SerhiiKirik, line 33: so, when a component will Unmount - you add two more listeners?)
just use `removeEventListener`

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

src/App.tsx
```diff
@@ -7,33 +8,59 @@ function getRandomName(): string {
...
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+    document.addEventListener('contextmenu', this.handleRightClick);
+    document.addEventListener('click', this.handleLeftClick);
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentWillUnmount(): void {
```
SerhiiKirik, line 33: So, in the Clock component, you remove this type, but here - leave, why?
```suggestion
  componentWillUnmount() {
```

PR #1470, opened 2023-04-17
add slution

Pozdnya:
[DEMO LINK](https://pozdnya.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock/CLock';
+
+interface State {
+  clockName: string;
+  hasClock: boolean;
+}
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


vksazonov (approved):
Good Job!

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
+
+  timerId = 0;
```
liliia-livandovskaya, line 22:
```suggestion
  timerId: number | null = null;
```

src/components/Clock/CLock.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  componentWillUnmount(): void {
+    window.clearInterval(this.timerId);
+  }
+
+  render() {
+    const { clockName } = this.props;
+    const { today } = this.state;
```
liliia-livandovskaya, line 42:
```suggestion
    const { clockName, today } = this.state;
```

src/components/Clock/CLock.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {formatedDate(today)}
+        </span>
```
liliia-livandovskaya, line 54: you can use interpolation instead

liliia-livandovskaya (changes requested):
Great job! Lets work on comments

src/App.tsx
Pozdnya: Done

src/components/Clock/CLock.tsx
Pozdnya: I move out repeateblee code in function to another component, if I don't mind they call helper. So, where should I use interpolation?

src/components/Clock/CLock.tsx
Pozdnya: How can I use `clockName` from the state, if I get it from props?

PR #1480, opened 2023-04-17
solution

artem-kyrylenko:
[DEMO LINK](https://artem-kyrylenko.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,92 @@
...
+  hasClock: boolean;
+}
+
+export class App extends Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
+
+  timerId: number | null = null;
```
Olekk17, line 22: You can just set this value to 0 and do not make it such complicated with null

src/App.tsx
```diff
@@ -1,39 +1,92 @@
...
+
+//         {' time is '}
+
+//         <span className="Clock__time">
+//           {today.toUTCString().slice(-12, -4)}
+//         </span>
+//       </div>
+//     </div>
+//   );
+// };
```
Olekk17, line 92: You should remove all unused commented code from commit

src/App.tsx
SerhiiKirik: null - it's ok if you don't have a value yet

src/App.tsx
```diff
@@ -1,39 +1,92 @@
...
+        {hasClock && <Clock clockName={clockName} />}
       </div>
-    </div>
-  );
-};
+    );
+  }
+}
+
+// export const App: React.FC = () => {
```
SerhiiKirik, line 63: Remove all comments

SerhiiKirik (changes requested):
Great 🔥 
Let's remove all comments 🧑‍💻

src/App.tsx
artem-kyrylenko: Done)

SerhiiKirik (approved):
Great 🔥

PR #1482, opened 2023-04-17
Solution

vlad-ptashnichenko:
[DEMO LINK](https://vlad-ptashnichenko.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,47 @@
...
+    }, 3300);
+
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
+
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
```
etojeDenys, line 34: This event listener should be removed on componentWillUnmount to prevent memory leaks 🧹

src/App.tsx
```diff
@@ -1,39 +1,47 @@
...
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  componentDidMount() {
+    window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+
```
etojeDenys, line 26: You should clear the interval on componentWillUnmount to prevent memory leaks 🧹

src/App.tsx
```diff
@@ -1,39 +1,47 @@
...
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
```
etojeDenys, line 30: This event listener should be removed on componentWillUnmount to prevent memory leaks 🧹

etojeDenys (changes requested):

PR #1486, opened 2023-04-18
solution

danylo-dobrovolskyi:
[DEMO LINK](https://danylo-dobrovolskyi.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+  }
+
+  RightClick = (event: MouseEvent) => {
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
+
+  LeftClick = () => {
+    this.setState({ hasClock: true });
+  };
```
kukla1989, line 47: you should rename names of this methods to something like this: "handleLeftClick" according to checklist:
3. [CODE STYLE] - Method's name should start with a verb (so you could easily tell what your method actually do)

kukla1989 (changes requested):
good job!!!

src/App.tsx
danylo-dobrovolskyi: good suggestion, thanks)

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+}
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+export class App extends Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+    timerId: 0,
```
SerhiiKirik, line 21: Would be better if you move this timer from the state, bc you trigger this state each 3s

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    const { hasClock, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {hasClock && <Clock name={clockName} />}
```
SerhiiKirik, line 55:
```suggestion
        <h1>React clock</h1>
        
        {hasClock && <Clock name={clockName} />}
```

SerhiiKirik (approved):
Great 🔥

PR #1491, opened 2023-04-18
clok solution

VladislavaT:
[DEMO LINK](https://VladislavaT.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './component';
+
+interface State {
+  hasClock: boolean,
+  clockName: string,
+}
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/component/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (this.props.clockName !== prevProps.clockName) {
+      window.console.debug(`Renamed from ${prevProps.clockName} to ${this.props.clockName}}`);
+    }
+
+    if (this.state.today !== prevState.today) {
+      window.console.info(this.state.today);
+    }
```
Hamuud, line 33: You can just put console.info inside ComponentDidMount, and it will be work the same way

src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount() {
+    document.removeEventListener('contextmenu', this.handleRightClick);
+    document.removeEventListener('click', this.handleLeftClick);
+    if (this.timerId) {
+      window.clearInterval(this.timerId);
+    }
```
Hamuud, line 38: Also you no need you conditional there
```suggestion
      window.clearInterval(this.timerId);
```

Hamuud (changes requested):
Good job, pay attention to comments🙃

I hope that the future of your life will be so carefree that the watch⏰ will be the most unnecessary item in your life

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends Component<{}, State> {
+  state: State = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  timerId = 0;
```
jstmpelowycz, line 22: would prefer `timerId` to be `null` initially

jstmpelowycz (approved):

PR #1492, opened 2023-04-18
WorkedVersion

AlexRog01:
[DEMO LINK](https://AlexRog01.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+
+import { Clock } from './components/clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+  clockName: string;
+};
+
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  righClickHandler = (event: MouseEvent): void => {
```
jstmpelowycz, line 23:
```suggestion
  handleRightClick = (event: MouseEvent): void => {
```
and so on

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
-  // this code stops the timer
-  window.clearInterval(timerId);
+  nameInteralHandler = () => {
+    this.setState({ clockName: getRandomName() });
+  };
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  nameTimerId = 0;
```
jstmpelowycz, line 37: better to set number or null

jstmpelowycz (approved):

PR #1498, opened 2023-04-18
solution

yapelenua:
[DEMO LINK](https://yapelenua.github.io/react_clock/

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "mochawesome": "^7.1.3",
     "mochawesome-merge": "^4.2.0",
     "mochawesome-report-generator": "^6.2.0",
-    "node-sass": "^6.0.1",
+    "sass": "*",
     "stylelint": "^13.13.1",
     "typescript": "^4.6.3"
   },
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "mochawesome": "^7.1.3",
     "mochawesome-merge": "^4.2.0",
     "mochawesome-report-generator": "^6.2.0",
-    "node-sass": "^6.0.1",
+    "sass": "*",
```
SerhiiKirik, line 33: In the future - Do not add extra files to PR
To add the file to a commit use `git add src/.` & `git commit -m 'your message'`

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
-        </span>
+export class App extends Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  timerId = 0;
+
+  componentDidMount(): void {
```
SerhiiKirik, line 24: It's extra, bc TS already knows about React types
```suggestion
  componentDidMount() {
```

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+    document.addEventListener('click', this.handleLeftClick);
```
SerhiiKirik, line 28:
```suggestion
    }, 3300);
    
    document.addEventListener('click', this.handleLeftClick);
```

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+    document.addEventListener('contextmenu', this.handleRightClick);
+  }
+
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+    document.addEventListener('click', this.handleLeftClick);
```
SerhiiKirik, line 34: So, after the component will be removed - you add one move `event`?
Just use `removeEventListener`

SerhiiKirik (changes requested):
Well done 🔥 
Let's make some improvements 🧑‍💻

src/App.tsx
```diff
@@ -1,39 +1,61 @@
-import React from 'react';
```
jstmpelowycz, line 1: Instead of importing `Component` separately, you can import it along with `React`. For example, `import React, { Component } from 'react'`.

src/Components/Clock.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return date.toUTCString().slice(-12, -4);
+}
+
+interface Props {
+  clockName: string;
+}
+
+interface State {
+  today: string,
```
jstmpelowycz, line 14: choost a better name

src/Components/Clock.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+
+function createTime() {
```
jstmpelowycz, line 3:
```suggestion
function getCurrentTime() {
```

src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
+  clockName: string,
+}
+
+export class App extends Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  timerId = 0;
```
jstmpelowycz, line 22: better to set number or null

jstmpelowycz (approved):

PR #1503, opened 2023-04-25
Develop

LBondarchuk:
[DEMO LINK](https://LBondarchuk.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,64 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string
+};
+
+function getRandomName() {
+  const value = Date.now().toString().slice(-4);
+
+  return `Clock-${value}`;
+}
+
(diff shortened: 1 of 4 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,64 @@
...
+function getRandomName() {
+  const value = Date.now().toString().slice(-4);
+
+  return `Clock-${value}`;
+}
+
+export class Clock extends Component<Props> {
+  state = {
+    today: new Date(),
+    clockName: this.props.clockName,
```
DarkMistyRoom, line 16: you should use only one source of clockName. It already exists in the App. Use it only from the props and change the name there

DarkMistyRoom (changes requested):
proofread [the checklist](https://github.com/mate-academy/react_clock/blob/master/checklist.md)

PR #1509, opened 2023-04-27
React clock task was done

SamirMamedov1010:
[DEMO LINK](https://SamirMamedov1010.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,61 @@
+import { Component } from 'react';
+
+type Props = {
+  clockName: string,
+};
+
+type State = {
+  today: Date,
+};
+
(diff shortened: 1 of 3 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const { clockName } = this.props;
+
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
```
IvanFesenko, line 54: You can wrap in by span too
```suggestion
        <span> time is </span>
```

IvanFesenko (approved):
Well done

PR #1519, opened 2023-05-05
solution

MikiOOOO:
https://MikiOOOO.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


sylcym (approved):

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+type AppState = {
```
d-wasilewski, line 11: use interface to describe objects

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
-          {clockName}
-        </strong>
+  componentDidUpdate(
+    _prevProps: Readonly<{}>, prevState: Readonly<AppState>,
+  ): void {
+    if (prevState.clockName !== this.state.clockName && this.state.hasClock) {
+      // eslint-disable-next-line
+      console.debug(`Renamed from ${prevState.clockName} to ${this.state.clockName}`);
+    }
+  }
```
d-wasilewski, line 50: Maybe do it in the Clock component? You pass the name there anyway, and you would need to only use prevProps there

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1520, opened 2023-05-05
Develop

Ilia991:
https://Ilia991.github.io/react_clock/

### DIFF
--- a/src/component/Clock/Clock.tsx
+++ b/src/component/Clock/Clock.tsx
@@ -1,0 +1,50 @@
+import React from 'react';
+
+type ClockType = {
+  clockname:string;
+};
+
+type State = {
+  today: Date;
+};
+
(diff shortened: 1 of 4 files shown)


src/component/Clock/Clock.tsx
```diff
@@ -0,0 +1,50 @@
...
+type State = {
+  today: Date;
+};
+
+export class Clock extends React.Component<ClockType, State> {
+  state:State = {
+    today: new Date(),
+  };
+
+  upDate: NodeJS.Timer | null = null;
```
etojeDenys, line 16: Use `number` type instead of `NodeJS.Timer` for the `upDate` variable, as `setInterval` returns a number in the browser environment.

etojeDenys (approved):

PR #1527, opened 2023-05-08
solution

kq333:
[DEMO LINK](https://kq333.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+
+interface ClockProps {
+  name: string;
+}
+
+interface ClockState {
+  time: string;
+  hasClock: boolean;
+}
+
+const getDate = () => new Date().toUTCString().slice(-12, -4);
+
+export class Clock extends React.Component<ClockProps, ClockState> {
+  private intervalId?: number;
+
+  state: ClockState = {
+    time: getDate(),
+    hasClock: true,
+  };
+
+  componentDidMount() {
+    this.intervalId = window.setInterval(() => {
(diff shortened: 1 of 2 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  hideClock = (event: MouseEvent) => {
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
+
+  showClock = (event: MouseEvent) => {
+    event.preventDefault();
+    this.setState({ hasClock: true });
+  };
```
d-wasilewski, line 47: they are not needed here

src/components/Clock.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const { name } = this.props;
+
+    return (
+      <div className="Clock">
+        {hasClock && (
+          <>
+            <strong className="Clock__name">
+              { name}
+            </strong>
+            {}
```
d-wasilewski, line 60:
```suggestion
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const getDate = () => new Date().toUTCString().slice(-12, -4);
+
+export class Clock extends React.Component<ClockProps, ClockState> {
+  private intervalId?: number;
+
+  state: ClockState = {
+    time: getDate(),
+    hasClock: true,
```
d-wasilewski, line 19: you shouldn't have this variable here, it should be only in the App and render entire component based on it

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1529, opened 2023-05-08
Solution

KielpinskiJ:
[DEMO LINK](https://KielpinskiJ.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
+  clockName: string;
+}
+
+export class App extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  intervalId: number | undefined;
```
d-wasilewski, line 24: You can give this a default value of 0 and keep it always as a number
```suggestion
  intervalId = 0;
```

d-wasilewski (changes requested):

d-wasilewski (changes requested):
remember to change accordingly all similar occurences after CR, in this case you have timerId both in Clock and App

kasiakubisiak (approved):

PR #1534, opened 2023-05-09
Solution

dianaavem:
[DEMO LINK](https://dianaavem-a.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,36 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
-function getRandomName(): string {
-  const value = Date.now().toString().slice(-4);
-
-  return `Clock-${value}`;
-}
-
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
-
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
-
-  // this code stops the timer
-  window.clearInterval(timerId);
-
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
-
-        {' time is '}
-
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
-      </div>
-    </div>
-  );
+type State = {
+  isContextmenu: boolean,
 };
+
+export class App extends Component<{}, State> {
+  state = {
+    isContextmenu: true,
+  };
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,36 @@
...
+  componentDidMount() {
+    document.addEventListener('contextmenu', (e) => {
+      e.preventDefault();
+      this.setState({ isContextmenu: false });
+    });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => {
```
tetiana-ivanchenko, line 20: you should remove lesteners on unmount

tetiana-ivanchenko (changes requested):
fix your demo link

src/App.tsx
```diff
@@ -1,39 +1,44 @@
...
-  const value = Date.now().toString().slice(-4);
+type State = {
+  isContextmenu: boolean,
+};
 
-  return `Clock-${value}`;
-}
+export class App extends Component<{}, State> {
+  state = {
+    isContextmenu: true,
```
DarkMistyRoom, line 11:
```suggestion
    hasClock: true,
```

DarkMistyRoom: image

src/App.tsx
```diff
@@ -1,39 +1,44 @@
...
 
-        {' time is '}
+    return (
+      <div className="App">
+        <h1>React clock</h1>
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+        {isContextmenu && <Clock />}
```
DarkMistyRoom, line 40: Clock should receive the name
image

DarkMistyRoom (changes requested):
- Please implement this requirement

image

- there are no logs about renaming
image

src/components/Clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+type Props = {
+  name: string,
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    date: new Date(),
+    isMounted: false,
```
SanyaBratashchuk, line 15: why do you need this in state?

src/components/Clock.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    this.timerId = window.setInterval(() => {
+      const date = new Date();
+
+      this.setState({ date });
+
+      if (this.state.isMounted) {
+        // eslint-disable-next-line
+        console.info(date);
+      }
```
SanyaBratashchuk, line 31: Why do you need to check it inside componentDidMount method?)

SanyaBratashchuk (changes requested):

OleziO (approved):

PR #1535, opened 2023-05-09
solution v1

bjeziorek:
[DEMO LINK](https://bjeziorek.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+  }
+
+  toggleClock = (event:Event) => {
+    event.preventDefault();
+    const isShown = this.state.isClockShown;
+
+    this.setState({
+      isClockShown: !isShown,
+    });
+  };
```
d-wasilewski, line 50: it doesn't work exactly as intented with this toggle function and always changing the `isClockShown` state to the opposite, because for now you can just keep clicking with either left or right click and the clock will aways toggle. The goal is to change the state to true or false depending on whether right or left mouse button has been clicked. For that you need to create 2 separate functions  with slightly different purpose

src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface Props {
+  name: string,
+}
+
+export class Clock extends React.Component <Props, State> {
+  state:Readonly<State> = {
+    date: '',
+  };
+
+  today = new Date().toUTCString().slice(-12, -4) ;
```
d-wasilewski, line 16: do you ever use this variable?

d-wasilewski: consider creating a function to return you this date `new Date().toUTCString().slice(-12, -4)`

d-wasilewski (changes requested):

PR #1536, opened 2023-05-09
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_clock/

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,67 @@
+import { Component } from 'react';
+
+type Props = {
+  name: string;
+};
+
+type State = {
+  clockName: string,
+  today: Date,
+};
+
+function getRandomName(): string {
+  const value = Date.now().toString().slice(-4);
+
+  return `Clock-${value}`;
+}
+
+export class Clock extends Component<Props, State> {
+  state = {
+    clockName: this.props.name,
+    today: new Date(),
+  };
+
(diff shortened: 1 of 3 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const { clockName, today } = this.state;
+
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
```
OleziO, line 65: Remove empty lines and wrap text with span

OleziO (changes requested):
Fix formatting, log the time and fix the delay of date showing

src/Clock.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+        {' time is '}
+        <span className="Clock__time">
+          {today.toUTCString().slice(-12, -4)}
+        </span>
```
OleziO, line 66: Combine it and use template strings.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO (changes requested):

witflash (approved):
Good job! 👍

PR #1541, opened 2023-05-11
initial version

Ozwiena:
[DEMO LINK](https://ozwiena.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
+
+interface State {
+  clockName: string,
+  hasClock: boolean,
+}
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+    });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: true });
+    });
+  }
```
d-wasilewski, line 38: create functions for this code, and use it in both and and remove listeners
```suggestion
    document.addEventListener('click', this.handleClick)
  }
```

d-wasilewski: you don't need to prevent default in 'click' event

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1548, opened 2023-05-18
Clock wsa created

Pavlo-Petrashevskyi:
- [DEMO LINK](https://Pavlo-Petrashevskyi.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,53 @@
+import { Component } from 'react';
+
+type ClockProps = {
+  clockName: string;
+};
+
+type ClockStates = {
+  today: Date;
+};
+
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  render() {
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.clockName}
+        </strong>
+
+        {' time is '}
+
```
OleziO, line 46: Remove empty lines and wrap it with span

OleziO (approved):

src/components/Clock.tsx
Pavlo-Petrashevskyi: Done

PR #1549, opened 2023-05-19
Task8

Ryshon:
[DEMO LINK](https://Ryshon.github.io/react_clock/)


anstsot (changes requested):
Looks good from a code perspective but your demo link isn’t working. could you check it, please?

PR #1561, opened 2023-05-29
done

generatedman:
[DEMO LINK](https://generatedman.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
+
+type State = {
+  hasClock: boolean;
+  clockName: string;
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


dinaminahi (changes requested):
Good job 👍 
But there are few moments we need to fix

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+
+  handleClick = () => {
+    this.setState({ hasClock: true });
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {this.state.hasClock && <Clock name={this.state.clockName} /> }
```
dinaminahi, line 54: Also don't forget that we can destructure from state and props
```suggestion
  render() {
  const { hasClock, clockName } = this.state;
    return (
      <div className="App">
        <h1>React clock</h1>
        {hasClock && <Clock name={clockName} /> }
```

dinaminahi (approved):
Well done 💪

PR #1562, opened 2023-05-30
add task solution

NickMitchell61:
[DEMO LINK](https://NickMitchell61.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
```
33Orange, line 32: Create separate function for event callback.

```suggestion
    document.addEventListener('contextmenu', this.onContextmenu);
```

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+  }
```
33Orange, line 37: Same here
```suggestion
    document.addEventListener('click', this.onClickDocument);
  }
```

src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+  }
```
33Orange, line 41: Don't forget to clear all listeners on unmount.
It is not necessary to do this in this situation, but it is good practice
```suggestion
  componentWillUnmount() {
    window.clearInterval(this.timerId);
    document.removeEventListener('click', this.onClickDocument);
  }
```

33Orange (changes requested):
Well done.

A few moments

src/App.tsx
```diff
@@ -1,39 +1,63 @@
...
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  onClickDocument = () => {
+    this.setState({ hasClock: true });
+  };
 
-        {' time is '}
+  onContextMenu = (event: { preventDefault: () => void; }) => {
```
silchyn, line 44: investigate how to set the type for this callback properly

silchyn (changes requested):

PR #1569, opened 2023-06-04
version

Grechana:
https://Grechana.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,12 @@
-import React from 'react';
+import { Component } from 'react';
+import { Clock } from './Clock';
 import './App.scss';
+
+type State = {
+  hasClock: boolean,
+  clockName: string,
+  previousClockName: string,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+      }));
+    }, 3300);
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount() {
+    clearInterval(this.timerForClock);
+  }
```
polosanya, line 42: Remove event listeners for 'contextmenu' and 'click' events on componentWillUnmount.

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+    previousClockName: '',
```
polosanya, line 21: There is no need to add previousClockName to the state. Think the other ways of how can you get it

src/Clock.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  render() {
+    const {
+      today,
+    } = this.state;
+
+    const {
+      clockName,
+    } = this.props;
```
polosanya, line 48:
```suggestion
    const { today } = this.state;
    const { clockName } = this.props;
```

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+      // eslint-disable-next-line no-console
+      console.info(this.state.today.toUTCString().slice(-12, -4));
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    if (prevProps.clockName !== this.props.clockName) {
+      // eslint-disable-next-line no-console
+      console.debug(`Renamed from ${this.props.previousClockName} to ${this.props.clockName}`);
+    }
```
agniya-i, line 33: You can use `prevProps.clockName` instead of adding `previousClockName` to state

src/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export class Clock extends Component<Props, State> {
+  state = {
+    today: new Date(),
+  };
+
+  timerForDate = 0;
+
+  timerForClockinfo = 0;
```
agniya-i, line 19: Remove `timerForClockinfo` if you don't use it

agniya-i (changes requested):

PR #1581, opened 2023-06-09
Add task solution

IrynaBalandiukh:
[DEMO LINK](https://IrynaBalandiukh.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,57 @@
+/* eslint-disable no-console */
+import React from 'react';
+
+type Props = {
+  clockName: string;
+};
+
+type State = {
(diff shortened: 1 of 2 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+type Props = {
+  clockName: string;
+};
+
+type State = {
+  today: Date;
+};
+
+export class Clock extends React.Component<Props, State> {
+  state:State = {
```
dinaminahi, line 13: You already set the type of state in line before, so no need to do it here also

dinaminahi (approved):
Well done 👍

PR #1596, opened 2023-06-11
add task solution

EugeneSherepa:
[DEMO LINK](https://EugeneSherepa.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,30 +26,28 @@ export class App extends React.Component<{}, State> {
       this.setState({ clockName: getRandomName() });
     }, 3300);
 
-    document.addEventListener('click', () => {
-      this.setState({ hasClock: true });
-    });
+    document.addEventListener('click', this.handleRightClick);
 
-    document.addEventListener('contextmenu', (event) => {
-      event.preventDefault();
-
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -7,33 +8,60 @@ function getRandomName(): string {
...
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+    document.removeEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
```
33Orange, line 45: Are you sure this will work?

Create handlers for event callbacks

src/App.tsx
```diff
@@ -7,33 +8,60 @@ function getRandomName(): string {
...
 
-        {' time is '}
+    document.removeEventListener('contextmenu', (event) => {
+      event.preventDefault();
```
33Orange, line 48: Same here

33Orange (changes requested):
Almost done

Create handlers for event callbacks

naziamoff (approved):

PR #1597, opened 2023-06-11
add task solution

hekaq:
[DEMO LINK](https://hekaq.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './components/clock/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+  }
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  componentWillUnmount(): void {
```
dinaminahi, line 40: Checklist p.1 
image

src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+    document.addEventListener('contextmenu', (event) => {
```
dinaminahi, line 29: Handle callbacks to separate functions

dinaminahi (changes requested):
Good job 👍 
Few things to fix

dinaminahi (approved):
Well done 👍 

But it is better to be consistent: in first component you added `void` returned type to lifecycle methods, but in the second you don't have this
Either add to all, or remove it from all

PR #1599, opened 2023-06-11
Solution

bohdan-mykhailenko:
[DEMO LINK](https://bohdan-mykhailenko.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,11 +2,59 @@ import React from 'react';
 import './App.scss';
 import { Clock } from './Clock';
 
-export const App: React.FC = () => {
-  return (
-    <div className="App">
-      <h1>React clock</h1>
-      <Clock />
-    </div>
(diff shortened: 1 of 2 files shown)


naziamoff (changes requested):
demo does not work as expected, nothing changes when I click

bohdan-mykhailenko:
> demo does not work as expected, nothing changes when I click

I've redeployed. It should work

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
+import { Clock } from './Clock';
 
-function getRandomName(): string {
+const getRandomName = (): string => {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
+};
+
+interface AppState {
```
naziamoff, line 11:
```suggestion
interface State {
```

PR #1603, opened 2023-06-12
Clock class created

olehmahera:
[DEMO LINK](https://olehmahera.github.io/react_clock/)

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,58 @@
+import { Component } from 'react';
+
+type Props = {
+  name: string;
+};
+
+type State = {
+  currentTime: Date
+};
+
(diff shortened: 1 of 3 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {currentTime.toUTCString().slice(-12, -4)}
```
dinaminahi, line 53: It is better to move  calculations or formatting to a variable or function from html

dinaminahi (approved):
Well done 👍

PR #1606, opened 2023-06-12
add clock solution

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_clock/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Create a `Clock` class component that will update the time every second using a 
     Renamed from oldName to newName
     ```
 - to see `console.debug` messages enable the `verbose` level in DevTools console:
-    
+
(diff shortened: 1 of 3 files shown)


dinaminahi (changes requested):
Some files are missing 😞

src/components/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+type State = {
+  today: Date,
+};
+
+type Props = {
+  name: string,
+};
+
+function toNormalDate(date: Date) {
```
naziamoff, line 11: function names should start with a verb
```suggestion
function normalizeDate(date: Date) {
```

naziamoff (approved):
looks good 👍

PR #1620, opened 2023-06-13
task was done

anton-chornobai:


anton-chornobai:
[DEMO LINK](https://anton-chornobai.github.io/react_clock/)

anton-chornobai:
Locally all tests passed i dont know  what can be wrong...

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+    document.addEventListener('click', this.showClock);
+  }
+
+  componentWillUnmount() {
+    this.setState({ hasClock: false });
+    document.removeEventListener('contextmenu', this.hideClock);
+    document.removeEventListener('click', this.showClock);
+  }
+
+  showClock = (event: Event) => {
```
etojeDenys, line 39: There is a minor issue with the event type. Change the type of 'event' to 'MouseEvent' instead of 'Event'.

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+
+    document.addEventListener('contextmenu', this.hideClock);
+    document.addEventListener('click', this.showClock);
+  }
+
+  componentWillUnmount() {
+    this.setState({ hasClock: false });
```
etojeDenys, line 34: You should clear the interval in componentWillUnmount instead of changing the state. Replace this line with `window.clearInterval(this.timerId);`.

src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+    document.removeEventListener('contextmenu', this.hideClock);
+    document.removeEventListener('click', this.showClock);
+  }
+
+  showClock = (event: Event) => {
+    event.preventDefault();
+    this.setState({ hasClock: true });
+  };
+
+  hideClock = (event: Event) => {
```
etojeDenys, line 44: There is a minor issue with the event type. Change the type of 'event' to 'MouseEvent' instead of 'Event'.

etojeDenys (changes requested):

etojeDenys:
also don't forget to add Clock file to your PR

src/components/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {today.toUTCString().slice(-12, -4)}
```
VitaliyBodnar, line 52: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

VitaliyBodnar (changes requested):
Almost done!

etojeDenys (approved):
looks good

PR #1629, opened 2023-06-17
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen
.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
+import { Clock } from './Clock';
+
+type State = {
+  hasClock: boolean,
+  clockName: string
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+    const { today } = this.state;
+    const { clockName } = this.props;
+
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
```
OleziO, line 51: Wrap with span and remove empty lines

OleziO (changes requested):

src/Clock.tsx
Kinderwagen: code stops working when changed
image_2023-06-18_12-02-53

src/App.tsx
```diff
@@ -1,39 +1,55 @@
...
+    const { hasClock, clockName } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {
+          hasClock === true
+            ? <Clock clockName={clockName} />
+            : null
+        }
```
Anastasiia-Svintsova, line 51: You don't need to use a ternary operator here 
```suggestion
        {hasClock && <Clock clockName={clockName} />}
```

src/App.tsx
```diff
@@ -1,39 +1,55 @@
...
+
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+
+    this.timerId = window.setInterval(() => {
+      this.setState({
+        clockName: getRandomName(),
+      });
+    }, 3300);
```
Anastasiia-Svintsova, line 38: When you have event listeners or timers set up in the componentDidMount lifecycle method, it's generally a good practice to clean up and remove them when the component is unmounted to avoid potential memory leaks and unexpected behavior.

Anastasiia-Svintsova (changes requested):
Check my comments, please. Let's add a few fixes to improve the code

etojeDenys (approved):
lgtm

PR #1652, opened 2023-07-11
Develop

cebanoleksandr:
- DEMO LINK - https://cebanoleksandr.github.io/react_clock/


src/App.tsx
```diff
@@ -1,39 +1,38 @@
...
+  };
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {this.state.hasClock && <Clock />}
```
anstsot, line 34: use destructuring

anstsot (changes requested):
Several conditions from the task are not met:
- show the Clock on a left mouse click in the document (click event)
- every second print the time in the DevTools using console.info method (not the console.log);
- each time the name is changed, the Clock must print a message with an old name and a new name using the console.debug method (use componentDidUpdate):

anstsot (changes requested):
The old bugs are still reproducible (previous comment).
maybe you forgot to redeploy?

And678 (approved):
Approve :checkered_flag:

PR #1654, opened 2023-07-15
added class component clock

JasterHim:
[DEMO LINK](https://JasterHim.github.io/react_clock/)

### DIFF
--- a/src/componets/Clock.tsx
+++ b/src/componets/Clock.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+
+type Props = {
+  name: string
+};
+
+type State = {
+  time: Date
+};
+
+export class Clock extends React.PureComponent<Props, State> {
+  state = {
+    time: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState((currentTime) => ({
+        ...currentTime,
+        time: new Date(),
+      }));
+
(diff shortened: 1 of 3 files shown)


src/componets/Clock.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {this.state.time.toUTCString().slice(-12, -4)}
```
VitaliyBodnar, line 53: please don't call a function from the interface because it will be executed every time when the state change, so make sense to move it to the variable

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

src/componets/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {time.toUTCString().slice(-12, -4)}
```
VitaliyBodnar, line 54: i mean to move time.toUTCString().slice(-12, -4) to the variable and use useMemo

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #1656, opened 2023-07-17
Solution

adv555:
**[DEMO LINK](https://adv555.github.io/react_clock/)**

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
+  hasClock: boolean;
+};
+
+export class App extends React.Component<{}, State> {
+  state:State = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
+
+  clockId = 0;
```
GGsmou, line 22: It may be better to give a more descriptive name, for example `clockNameTimerId`

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+    document.addEventListener('contextmenu', this.handleDocumentRightClick);
+    document.addEventListener('click', this.handleDocumentLeftClick);
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentWillUnmount(): void {
```
GGsmou, line 35: It's not necessary to clarify `void` return type

src/App.tsx
```diff
@@ -1,39 +1,68 @@
...
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+type State = {
+  clockName: string;
+  hasClock: boolean;
+};
+
+export class App extends React.Component<{}, State> {
+  state:State = {
```
anstsot, line 17:
```suggestion
  state = {
```

anstsot (approved):
Good work, approving:sparkles:

PR #1666, opened 2023-07-20
add task solution

NazarBaraban:
https://nazarbaraban.github.io/react_clock/


kochetov-yevhen-may29 (approved):
Good solution :)

I suppose you should use document.addEventListener/removeEventListener instead window.addEventListener/removeEventListener for "click" and "contextmenu"
Because it is document events, and it doesn't affect the browser window.

And678 (approved):
Great work, approve :checkered_flag:

PR #1678, opened 2023-07-21
solution

AnatoliiLykhohodenko:
[DEMO LINK](https://AnatoliiLykhohodenko.github.io/react_clock/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4374,9 +4374,9 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "1.2.3",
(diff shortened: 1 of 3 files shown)


anstsot (changes requested):
Good start:four_leaf_clover:
Several conditions from the task are not met:
- every second print the time in the DevTools using `console.info` method (not the `console.log`);
- each time the name is changed, the Clock must print a message with an old name and a new name using the `console.debug` method (use `componentDidUpdate`)

YegorKorenkov (changes requested):
loss of context in hideClock and showClock
image

YegorKorenkov (approved):
Good job!

PR #1687, opened 2023-07-21
Solution

outofs:
https://outofs.github.io/react_clock/


src/App.tsx
```diff
@@ -1,39 +1,59 @@
...
+
+  handleClick = () => {
+    this.setState({ clockIsVisible: true });
+  };
+
+  componentWillUnount(): void {
+    document.removeEventListener('contextmenu', this.handleContextmenu);
+    document.removeEventListener('click', this.handleClick);
+    window.clearInterval(this.timerId);
+  }
```
ovynarchuk, line 46: you don't have to use `componentWillUnmount` in `App` component

src/components/Clock.tsx
```diff
@@ -0,0 +1,52 @@
...
+      console.info(this.state.today.toUTCString().slice(-12, -4));
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Readonly<Props>): void {
+    if (prevProps.clockName !== this.props.clockName) {
+      // eslint-disable-next-line no-console
+      console.debug(`Renamed from ${prevProps.clockName} to ${this.props.clockName}`);
+    }
+  }
```
ovynarchuk, line 31: your code is not working because of this
image

ovynarchuk (changes requested):
good start, but check the comments, because it's not working like it's supposed to

And678 (approved):
Good work, approved :heavy_check_mark:

PR #1688, opened 2023-07-21
add task solution

klmyana:


src/App.tsx
```diff
@@ -1,39 +1,57 @@
...
+    document.addEventListener('contextmenu', this.onRightClick);
+  }
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  componentWillUnmount(): void {
+    window.clearInterval(this.timerId);
+  }
```
vlopatko, line 37: According to the task, the App component will not unmount by default and clockName should always change, so this lifecycle method doesn't need. 
However, you can write the logic for this case 👍🏻

vlopatko (approved):
Good job! All works!💯

src/App.tsx
vlopatko, line 37: According to the task, when the App component unmounts you should remove all listeners.
image

vlopatko, line 37:
```suggestion
  componentWillUnmount(): void {
    window.clearInterval(this.timerId);
    document.removeEventListener('click', this.handleLeftClick);
    document.removeEventListener('contextmenu', this.handleRightClick);
  }
```

vlopatko (changes requested):
Good job! But need to add some changes by the checklist.

PR #1689, opened 2023-07-21
solution

olegpaliy:
https://olegpaliy.github.io/react_clock/

### DIFF
--- a/src/components/Clock.tsx
+++ b/src/components/Clock.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+
+type Props = {
+  name: string;
+};
+
+type State = {
+  currentDay: Date;
+};
+
+export class Clock extends React.Component<Props, State> {
+  state: State = {
+    currentDay: new Date(),
+  };
+
+  timer = 0;
+
+  componentDidMount(): void {
+    this.timer = window.setInterval(() => {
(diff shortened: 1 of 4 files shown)


melnyksergey7:
Demo link does not work

src/components/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  componentDidMount(): void {
+    this.timer = window.setInterval(() => {
+      this.setState({ currentDay: new Date() });
+      // eslint-disable-next-line no-console
+      console.info(this.state.currentDay.toUTCString().slice(-12, -4));
+    }, 1000);
+  }
+
+  componentDidUpdate(
```
melnyksergey7, line 26: it's not necessary divide this string on the several strings

src/components/Clock.tsx
```diff
@@ -0,0 +1,54 @@
...
+    prevProps: Readonly<Props>,
+  ): void {
+    if (prevProps.name !== this.props.name) {
```
melnyksergey7, line 29: you can use destructuring in this case

melnyksergey7 (changes requested):
Good start, but fix problems

And678 (approved):
Great, approved :heavy_check_mark:

PR #1722, opened 2023-08-07
Add class components

AmadeuszAndroid:
[DEMO LINK](https://amadeuszandroid.github.io/react_clock/)


src/App.tsx
```diff
@@ -1,39 +1,54 @@
...
-          {clockName}
-        </strong>
-
-        {' time is '}
-
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+export class App extends React.Component {
+  state = {
```
Radoslaw-Czerniawski, line 12: Generally it's a good practice to type your state
```tsx
interface AppState {
  hasClock: boolean;
  clockName: string;
  clockNameInterval: number;
}

export class App extends React.Component<{}, AppState> {
  state: Readonly<AppState> = {
    hasClock: true,
    clockName: 'Clock-0',
    clockNameInterval: -1,
  }
```
ReadOnly generic type will signify the next developer that he is working with state and that he should not modify it explicitly without using setState, `this.state.hasClock = false` will then correctly throw a typescript error

src/Clock.tsx
```diff
@@ -0,0 +1,60 @@
...
+function getCurrentClockTime() {
+  return new Date().toUTCString().slice(-12, -4);
+}
+
+type ClockProps = {
+  name: string
+};
+
+export class Clock extends React.Component<ClockProps> {
+  state = {
```
Radoslaw-Czerniawski, line 12: Same as above

src/Clock.tsx
```diff
@@ -0,0 +1,60 @@
...
+        });
+      }, 1000),
+    });
+  }
+
+  componentDidUpdate(prevProps: Readonly<ClockProps>) {
+    const oldName = prevProps.name;
+    const newName = this.props.name;
+
+    if (oldName !== newName) {
```
Radoslaw-Czerniawski, line 35: There is no need for declaring two variables for the sake of renaming values, just
```ts
if (this.props.name !== prevProps.name) {
```

Radoslaw-Czerniawski (changes requested):
Looks very good, needs some minor changes :D

src/App.tsx
AmadeuszAndroid: I added readonly type.

src/Clock.tsx
AmadeuszAndroid: I added readonly type.

src/Clock.tsx
AmadeuszAndroid: I destructured props.

src/Clock.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  render() {
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.name}
+        </strong>
+
+        {' time is '}
+
```
OleziO, line 57: Remove empty lines and wrap it with span

AmadeuszAndroid: Can You describe, why it is needed? On master branch, we have exactly [this code with empty lines](https://github.com/mate-academy/react_clock/blob/master/src/App.tsx#L31).

darokrk (approved):
Looks good 💪🏽

PR #1723, opened 2023-08-07
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -29,10 +29,8 @@ export class Time extends React.Component<{}, TimeState> {
 
   timerId = 0;
 
-  timerId2 = 0;
-
   componentDidMount() {
-    this.timerId2 = window.setInterval(() => {
+    this.timerId = window.setInterval(() => {
       this.setState({ today: new Date() });
       console.info(this.state.today.toUTCString().slice(-12, -4));
     }, 1000);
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -7,33 +13,111 @@ function getRandomName(): string {
...
+    today: new Date(),
+  };
+
+  timerId = 0;
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+  timerId2 = 0;
```
darokrk, line 32: do you need this `timerId2` here? think using just `timerId` in this component should be enough

src/App.tsx
```diff
@@ -7,33 +13,111 @@ function getRandomName(): string {
...
+      console.info(this.state.today.toUTCString().slice(-12, -4));
+    }, 1000);
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+    window.clearInterval(this.timerId2);
```
darokrk, line 43: think this line is redundant

darokrk (changes requested):
Nicely done 👍🏽  few improvements are needed and we got this 🛠️ 🔥 !

darokrk (approved):
Good job 👍🏽 !

PR #1724, opened 2023-08-08
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_clock

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+
+type Props = {
+  name: string
+};
+
+type State = {
+  today: Date
+};
+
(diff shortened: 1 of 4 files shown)


src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type Props = {
+  name: string
+};
+
+type State = {
+  today: Date
+};
+
+export class Clock extends React.Component<Props, State> {
```
volodymyr-soltys97, line 11: The Clock component is not printing a message with the old and new name when the name is changed.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Well done!

PR #1725, opened 2023-08-08
Develop

domkab:
https://domkab.github.io/react_clock

### DIFF
--- a/src/components/Clock/Clock.tsx
+++ b/src/components/Clock/Clock.tsx
@@ -36,12 +36,16 @@ export class Clock extends React.Component<ClockProps, ClockState> {
   }
 
   componentDidUpdate(_prevProps: {}, prevState: ClockState) {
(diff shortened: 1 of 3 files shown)


src/components/Clock/Clock.tsx
```diff
@@ -0,0 +1,118 @@
...
+    this.timerID = setInterval(
+      () => this.updateTime(), 1000,
+    );
+
+    this.nameTimerID = setInterval(
+      () => this.updateName(), 3300,
+    );
+
+    document.addEventListener('contextmenu', this.handleRightClick);
+    document.addEventListener('click', this.handleLeftClick);
```
maxim2310, line 35: this func is working wrong, i can hidden and show clock only right mouse button, maybe you dont deploy your last version, because code looks right, check it please?

maxim2310 (changes requested):
Good work but check comments

PR #1728, opened 2023-08-16
add solution

denis-kononenko:
[DEMO LINK](https://denis-kononenko.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
+
+type State = {
+  clockName: string,
+  hasClock: boolean,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,39 +1,55 @@
...
+      this.setState({ hasClock: true });
+    });
+  }
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+  componentWillUnmount(): void {
+    window.clearInterval(this.timerId);
```
anastasiiavorobiova, line 41: Consider removing event listeners here to prevent memory leaks

anastasiiavorobiova (changes requested):
Fantastic work! Almost done!

src/App.tsx
denis-kononenko: ok, I will add, I thought about it, but I thought as it is one component, there aren't reasons to do it)

anastasiiavorobiova (approved):
Good work!

PR #1736, opened 2023-08-24
add clock

w315:
[DEMO LINK](https://w315.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Timer } from './components/Timer';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,69 @@
...
-  let clockName = 'Clock-0';
+interface State {
+  clockName: string;
+  hide: boolean;
+}
+
+export class App extends React.Component<{}, State> {
+  state = {
+    clockName: 'Clock-0',
+    hide: true,
```
lerastarynets, line 19: `hide` property is confusing, you actually set it to `true` when showing clock, and set it to `false` when hiding, consider renaming it to something like `showClock`

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1740, opened 2023-08-25
Develop

anpiet:
[DEMO LINK](https://anpiet.github.io/react_clock/)


src/App.tsx
```diff
@@ -1,39 +1,48 @@
...
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  render() {
+    const { clockName, hasClock } = this.state;
+
+    document.addEventListener('contextmenu', (event) => {
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
```
choeqq, line 33: this should be done in mount component lifecycle method

src/App.tsx
```diff
@@ -1,39 +1,48 @@
...
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
```
choeqq, line 37: this should be done in mount component lifecycle method

src/App.tsx
choeqq, line 33: remember to remove listener when component will unmount

src/App.tsx
choeqq, line 37: remember to remove listener when component will unmount

choeqq (changes requested):

klaudia-klis (changes requested):
Make suggested changes and it should be good 👍🏼

choeqq (approved):

PR #1742, opened 2023-08-25
FINALLY

Dolnys:
[DEMO LINK](https://dolnys.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,58 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
-function getRandomName(): string {
+export function getRandomName(): string {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,58 @@
...
+    });
+
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
+  }
+
+  componentWillUnmount() {
+    window.clearInterval(this.timerId);
+  }
```
choeqq, line 43: you should also remove both listeners on unmount

choeqq (changes requested):

src/App.tsx
Dolnys: You are right, done !👍

PR #1750, opened 2023-08-28
Solution

Vira-V:
[DEMO LINK](https://Vira-V.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,66 @@
 import React from 'react';
 import './App.scss';
 
+import { Clock } from './components/Clock';
+
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
-
-  // This code starts a timer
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
+    document.removeEventListener('click', this.showClock);
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  startTimer = () => {
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+  };
```
choeqq, line 43: it's already done in mount lifecycle method, please use intervals only in lifecycle methods

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  hideClock = (event: MouseEvent) => {
+    event.preventDefault();
```
choeqq, line 46: I don't think that's necessary 🤔

choeqq (approved):

choeqq (changes requested):
please fix your code according to comment above ⬆️

choeqq (approved):

PR #1755, opened 2023-08-28
init

MaksymKos:
https://maksymkos.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,43 @@
...
+type Props = {
+  clockName: string,
+  hasClock: boolean,
+};
+
+export class Clock extends React.Component<Props> {
+  state = {
+    time: new Date(),
+    timerId: 0,
+  };
```
Viktor-Kost, line 13: The timerId should not be in the state because it does not affect rendering.

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
+
+  render() {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {this.state.hasClock && (
+          <Clock
+            clockName={this.state.clockName}
+            hasClock={this.state.hasClock}
```
Viktor-Kost, line 60: The 'hasClock' prop is unnecessary in the 'Clock' component as the 'Clock' will only be rendered when 'hasClock' is true.

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends React.Component {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+    timerId: 0,
```
Viktor-Kost, line 16: The timerId should not be in the state because it does not affect rendering.

Viktor-Kost (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
+export class App extends React.Component {
+  state = {
+    clockName: 'Clock-0',
+    hasClock: true,
+  };
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentDidMount(): void {
+    const timerIdInternal = window.setInterval(() => {
```
IvanFesenko, line 21: No need in extra variable
```suggestion
    timerId = window.setInterval(() => {
```

src/App.tsx
```diff
@@ -1,39 +1,66 @@
...
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+      this.setState({ clockName: name });
+    }, 3300);
 
-        {' time is '}
+    timerId = timerIdInternal;
```
IvanFesenko, line 31:
```suggestion

```

src/Clock.tsx
```diff
@@ -0,0 +1,43 @@
...
+    time: new Date(),
+  };
+
+  componentDidMount(): void {
+    const timerIdInternal = window.setInterval(() => {
+      this.setState({ time: new Date() });
+      console.info(this.state.time.toUTCString().slice(-12, -4));
+    }, 1000);
+
+    timerId = timerIdInternal;
```
IvanFesenko, line 21: Same here

IvanFesenko (approved):

PR #1768, opened 2023-08-30
Clock class component

ilona-romaniuk:
https://ilona-romaniuk.github.io/react_clock/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,40 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
-function getRandomName(): string {
-  const value = Date.now().toString().slice(-4);
+export class App extends React.Component {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
 
-  return `Clock-${value}`;
-}
+  componentDidMount(): void {
+    document.addEventListener('contextmenu', (event: MouseEvent) => {
+      event.preventDefault();
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+      this.setState({ hasClock: false });
+    });
+  }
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,40 @@
...
+    });
+  }
+
+  componentDidUpdate(): void {
+    if (!this.state.hasClock) {
+      document.addEventListener('click', () => {
+        this.setState({ hasClock: true });
+      });
+    }
+  }
```
DarkMistyRoom, line 25: no need to add a new listener on each update. Do it once in the  componentDidMount

DarkMistyRoom: and remove listeners in the  `componentWillUnmount`

src/components/Clock.tsx
```diff
@@ -0,0 +1,76 @@
...
+      this.setState({ today: new Date() });
+
+      // eslint-disable-next-line no-console
+      console.info(this.state.today.toUTCString().slice(-12, -4));
+    }, 1000);
+
+    this.nameId = window.setInterval(() => {
+      this.name = this.state.clockName;
+      this.state.clockName = getRandomName();
+    }, 3300);
```
DarkMistyRoom, line 41: move name updating logic to the App to pass the tests

src/components/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  render() {
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.name}
+        </strong>
+
+        {' time is '}
```
OleziO, line 49: Wrap it with span.

OleziO (changes requested):

lerastarynets (approved):
You're doing brilliantly, keep it up!

PR #1772, opened 2023-09-01
solution

dseq0:
[DEMO LINK](https://dseq0.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/components/Clock.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  render() {
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.clockName}
+        </strong>
+
+        {' time is '}
```
OleziO, line 43: Wrap it with span.

OleziO (changes requested):

src/components/Clock.tsx
```diff
@@ -0,0 +1,51 @@
...
+    today: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ today: new Date() });
+
+      window.console.info(this.state.today.toUTCString().slice(-12, -4));
```
VitaliyBodnar, line 22:
```suggestion
      console.info(this.state.today.toUTCString().slice(-12, -4));
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }, 1000);
+  }
+
+  componentDidUpdate(prevProps: Props): void {
+    if (prevProps.clockName !== this.props.clockName) {
+      window.console.debug(`Renamed from ${prevProps.clockName} to ${this.props.clockName}`);
```
VitaliyBodnar, line 28:
```suggestion
      console.debug(`Renamed from ${prevProps.clockName} to ${this.props.clockName}`);
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {this.state.today.toUTCString().slice(-12, -4)}
```
VitaliyBodnar, line 46: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/App.tsx
```diff
@@ -1,39 +1,56 @@
...
+
+    document.addEventListener('click', () => {
+      if (!this.state.hasClock) {
+        this.setState({ hasClock: true });
+      }
+    });
+  }
+
+  componentWillUnmount(): void {
+    window.clearInterval(this.timerId);
```
VitaliyBodnar, line 45: you need to remove all event Listeners when component is destroyed

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):
Good job!

PR #1784, opened 2023-09-08
completed task react_clock

DenisChakhalian:
[DEMO LINK](https://denischakhalian.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,14 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Props = any;
+
+type State = {
+  hasClock: boolean;
+  clockName: string;
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,39 +1,72 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Props = any;
```
alexander-ignatow, line 6: Don't do so, unless you have a strong reason to.
By using `any` you effectively switch off all typescript checks

In your case please, look for alternatives:
1. Use `unknown` or `never`
2. Just remove typisation from `App` and let typescript to calculate all types for you.

src/App.tsx
```diff
@@ -1,39 +1,72 @@
...
+      this.setState({ clockName: getRandomName() });
+    }, 3300);
+  }
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentDidUpdate(_prevProps: Readonly<Props>, prevState: Readonly<State>) {
+    if (prevState.clockName !== this.state.clockName && this.state.hasClock) {
+      // eslint-disable-next-line no-console
+      console.debug(`Renamed from ${prevState.clockName} to ${this.state.clockName}`);
```
alexander-ignatow, line 39: It is ok for debug, but should be removed from "production" version

src/components/Clock.tsx
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      const newValue = new Date();
+
+      // eslint-disable-next-line no-console
+      console.info(newValue.toUTCString().slice(-12, -4));
```
alexander-ignatow, line 23: Per task:
> the time should not be printed to the console when the Clock is hidden (componentWillUnmount);

And you are doing right opposite :)

alexander-ignatow (changes requested):
Good, still some changes are needed

src/components/Clock.tsx
alexander-ignatow: My bad - your solution is right. Sory for confusion :)

Moroz-Dmytro (changes requested):

PR #1792, opened 2023-09-11
Added React_clock solution

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,63 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Components/Clock/Clock';
+
+interface State {
+  hasClock: boolean,
+  clockName: number,
+}
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
 
-  return `Clock-${value}`;
+  return value;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+export class App extends React.Component<{}, State> {
+  state: State = {
+    hasClock: true,
+    clockName: 0,
+  };
 
-  // This code starts a timer
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+  timerId = 0;
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: +getRandomName() });
+    }, 3300);
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('contextmenu', this.handleRightClick);
 
-      <div className="Clock">
-        <strong className="Clock__name">
-          {clockName}
-        </strong>
+    document.addEventListener('click', this.hadleLeftClick);
+  }
 
-        {' time is '}
+  componentWillUnmount(): void {
+    window.clearInterval(this.timerId);
+  }
 
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+  handleRightClick = (event: MouseEvent) => {
+    event.preventDefault();
+    this.setState({ hasClock: false });
+  };
+
+  hadleLeftClick = () => {
+    this.setState({ hasClock: true });
+  };
+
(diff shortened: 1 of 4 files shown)


AntonHerasymov (approved):
Good job!!!!

src/App.tsx
```diff
@@ -1,39 +1,63 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Components/Clock/Clock';
+
+interface State {
+  hasClock: boolean,
+  clockName: number,
+}
 
 function getRandomName(): string {
```
roman-mirzoian, line 10: In this function, you do not return a whole name, but only a random value
```suggestion
function getRandomValue(): string {
```

src/App.tsx
```diff
@@ -1,39 +1,63 @@
...
+  state: State = {
+    hasClock: true,
+    clockName: 0,
+  };
+
+  timerId = 0;
+
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: +getRandomName() });
```
roman-mirzoian, line 26: It is better to transfer the conversion to the number (+) type in the middle of the helper function.

src/Components/Clock/Clock.tsx
```diff
@@ -0,0 +1,68 @@
...
+  hasClock: boolean
+}
+
+interface State {
+  today: Date,
+}
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    today: new Date(),
```
roman-mirzoian, line 14:
```suggestion
    currentDate: new Date(),
```

src/Components/Clock/Clock.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {`Clock-${name}`}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {today.toUTCString().slice(-12, -4)}
```
roman-mirzoian, line 63: You do such transformations several times, so it's better to make it a helper function

roman-mirzoian (changes requested):

src/App.tsx
```diff
@@ -1,39 +1,63 @@
...
-  // this code stops the timer
-  window.clearInterval(timerId);
+    document.addEventListener('click', this.hadleLeftClick);
+  }
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+  componentWillUnmount(): void {
+    window.clearInterval(this.timerId);
```
Moroz-Dmytro, line 35: You need also to remove your eventListeners here

src/App.tsx
```diff
@@ -1,39 +1,63 @@
...
+    const { hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && (
+          <Clock
+            name={this.state.clockName}
+            hasClock={this.state.hasClock}
```
Moroz-Dmytro, line 57: Why you need to pass this prop? It will be always `true` because you render this component only if it is `true`

src/App.tsx
```diff
@@ -1,39 +1,63 @@
...
+  render() {
+    const { hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && (
+          <Clock
+            name={this.state.clockName}
```
Moroz-Dmytro, line 56: Use destructuring

src/Components/Clock/Clock.tsx
```diff
@@ -0,0 +1,69 @@
...
+      if (this.props.hasClock) {
+        this.timerId = window.setInterval(() => {
+          this.setState({ currentDate: new Date() });
+          // eslint-disable-next-line no-console
+          console.info(formatTime(this.state.currentDate));
+        }, 1000);
+      } else {
+        window.clearInterval(this.timerId);
+      }
+    }
```
Moroz-Dmytro, line 44: I guess you don't need this block of code

Moroz-Dmytro (changes requested):

PR #1794, opened 2023-09-12
Solution

Oleh0293:
[DEMO LINK](https://Oleh0293.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,11 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './Clock';
+
+type State = {
+  clockName: string,
+  hasClockName: boolean,
+};
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,60 @@
 import React from 'react';
 import './App.scss';
```
rostyslav48, line 2: It is a good practice to separate our components from other libraries by using 1 empty line

src/App.tsx
```diff
@@ -1,39 +1,60 @@
...
-
-        {' time is '}
-
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+export class App extends React.Component {
+  state: State = {
+    clockName: 'Clock-0',
+    hasClockName: true,
```
rostyslav48, line 19: Maybe it is better to rename this variable, because it deffinitely not about names) In the task it is suggested to name it `hasClock` (README, line 22)

src/Clock.tsx
```diff
@@ -0,0 +1,50 @@
...
+export class Clock extends React.Component<Props, State> {
+  state: State = {
+    nowDate: new Date(),
+  };
+
+  timerId = 0;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      const newValue = new Date();
```
rostyslav48, line 20: Do you really need to write a new Date into this constant? It is better to write it directly in the state, isn't it?
`this.setState({ nowDate: new Date() });`

src/Clock.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {nowDate.toUTCString().slice(-12, -4)}
```
rostyslav48, line 45: You are writing quite similar code in **line 23**, so, maybe it is better to create some method for it?

rostyslav48 (changes requested):
Well Done👍. I have written several comments to check. Your work has been rejected because of the last one, where you duplicated the code. Please fix it. All other comments are just my subjective point of how a good code should look like.

rostyslav48 (approved):
Now it looks awesome🤩)

src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
+
+  render() {
+    const { hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && (
+          <Clock clockName={this.state.clockName} />
```
Moroz-Dmytro, line 56: Also use destructuring

src/Clock.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+type State = {
+  nowDate: Date;
+};
+
+export class Clock extends React.Component<Props, State> {
+  static formatTime(date: Date): string {
+    return date.toUTCString().slice(-12, -4);
+  }
```
Moroz-Dmytro, line 14: Move this outside the component as a helper function

src/Clock.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  state: State = {
+    nowDate: new Date(),
```
Moroz-Dmytro, line 17:
```suggestion
    today: new Date(),
```

Moroz-Dmytro (changes requested):
You need to fix tests, they are failing

PR #1798, opened 2023-09-12
add task solution

ydratsevska:
[DEMO LINK](https://ydratsevska.github.io/react_clock/)


src/Components/Clock.tsx
```diff
@@ -0,0 +1,55 @@
...
+  today: Date,
+};
+
+type Props = {
+  clockName: string,
+};
+
+export class Clock extends React.Component<Props> {
+  state: State = {
+    today: new Date(),
```
roman-mirzoian, line 14:
```suggestion
    currentDate: new Date(),
```

src/Components/Clock.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {clockName}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {today.toUTCString().slice(-12, -4)}
```
roman-mirzoian, line 50: You make similar calculations twice, so you should make a separate helper function for this.

roman-mirzoian (changes requested):

PR #1803, opened 2023-09-12
implemented Clock class component

voievodik:
[DEMO LINK](https://voievodik.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,39 +1,62 @@
...
-  const timerId = window.setInterval(() => {
-    clockName = getRandomName();
-  }, 3300);
+type State = {
+  hasClock: boolean;
+  clockName: string;
+};
+
+export class App extends React.Component<Props, State> {
+  state: State = {
```
OlhaMomot, line 19: Maybe, if you don't use props in this component you can just put an empty object instead (<{}, State>), so you won't need special type for them as well.

OlhaMomot (approved):
Looks good & works as intended👍

src/components/Clock.tsx
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {this.props.name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {this.state.currentTime.toUTCString().slice(-12, -4)}
```
roman-mirzoian, line 47: It is better to put such calculations in separate variables to improve readability

roman-mirzoian (approved):
Well done 👍

PR #1805, opened 2023-09-12
created app class component & clock class component

polinavafik:
[DEMO LINK](https://polinavafik.github.io/react_clock/)


viktoria-roik (approved):
Good Job!)

src/components/Clock.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type State = {
+  today: string,
+};
+
+export class Clock extends React.Component<Props, State> {
+  timerId = 0;
+
+  state: State = {
+    today: new Date().toUTCString().slice(-12, -4),
```
roman-mirzoian, line 15: You save here exactly the time, not the entire date
```suggestion
    currentTime: new Date().toUTCString().slice(-12, -4),
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ today: new Date().toUTCString().slice(-12, -4) });
```
roman-mirzoian, line 20: You make similar calculations twice, so you should make a separate helper function for this.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #1822, opened 2023-09-13
add solution of react_clock task

vatatan:
[DEMO LINK](https://vatatan.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,39 +1,61 @@
 import React from 'react';
 import './App.scss';
 
+import { Clock } from './Clock';
+
+type State = {
+  clockName: string;
+  hasClock: boolean;
+};
+
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,39 +1,61 @@
...
+
+  render(): React.ReactNode {
+    const { hasClock } = this.state;
+
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+
+        {hasClock && (
+          <Clock name={this.state.clockName} />
```
roman-mirzoian, line 56: If you destructurize the state, the `clockName` can also be obtained by destructuring the state.

src/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return (
+      <div className="Clock">
+        <strong className="Clock__name">
+          {name}
+        </strong>
+
+        {' time is '}
+
+        <span className="Clock__time">
+          {date.toUTCString().slice(-12, -4)}
```
roman-mirzoian, line 52: You do a similar operation twice, so it's worth taking it out as a separate helper.

roman-mirzoian (changes requested):

PR #2290, opened 2024-05-14
add solution

stepanyukvova200:
[DEMO LINK](https://stepanyukvova200.github.io/react_clock/)


stepanyukvova200:
Also, if I decrease all expected times by 1, all tests will pass.

PR #2305, opened 2024-05-26
create a clock with classes components

Solovei-Andrii:


Solovei-Andrii:
https://Solovei-Andrii.github.io/react_clock/

PR #2735, opened 2025-03-20
add task solution

AnOvramenko:
[DEMO LINK](https://AnOvramenko.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -8,16 +8,20 @@ interface State {
   currTime: string;
 }
 
+const getCurrentTime = () => {
+  return new Date().toUTCString().slice(-12, -4);
+};
+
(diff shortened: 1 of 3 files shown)


src/Clock.tsx
```diff
@@ -0,0 +1,56 @@
...
+  name: string;
+}
+
+interface State {
+  currTime: string;
+}
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    currTime: new Date().toUTCString().slice(-12, -4),
```
daniil-trusov: It's better to create separate function that will return current time, because you dublicate this code on line 20;

daniil-trusov (approved):
Good job! Please check comment for details how you can improve your solution.

PR #2746, opened 2025-03-21
solution

ansofidev:
[DEMO LINK](https://sansamiste.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Clock } from './component/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,37 +1,55 @@
...
-        <span className="Clock__time">
-          {today.toUTCString().slice(-12, -4)}
-        </span>
+export class App extends React.Component<{}, State> {
+  state = {
+    hasClock: true,
+    clockName: 'Clock-0',
+  };
+
+  timerId: number = 0;
```
fromjkeee16: `0` is a valid value for `setInterval`, so if interval with id `0` already exists in the window, it can be stopped by a component if `window.clearInterval(this.timerId)` was, for example, in the `componentDidMount()` method
`window.clearInterval(undefined)` wont cause an error nor do anything
```suggestion
  timerId: number | undefined;
```

src/App.tsx
```diff
@@ -1,37 +1,55 @@
...
+    document.removeEventListener('click', this.handleLeftClick);
+    window.clearInterval(this.timerId);
+  }
+
+  render(): React.ReactNode {
+    return (
+      <div className="App">
+        <h1>React clock</h1>
+        {this.state.hasClock && <Clock clockName={this.state.clockName} />}
       </div>
```
fromjkeee16:
```suggestion
  render(): React.ReactNode {
   const { hasClock, clockName} = this.state;
   
    return (
      <div className="App">
        <h1>React clock</h1>
        {hasClock && <Clock clockName={clockName} />}
      </div>
```

src/component/Clock.tsx
```diff
@@ -0,0 +1,57 @@
...
+        `Renamed from ${prevProps.clockName} to ${this.props.clockName}`,
+      );
+    }
+  }
+
+  componentWillUnmount(): void {
+    if (this.timerId !== null) {
+      window.clearInterval(this.timerId);
+    }
+  }
```
fromjkeee16: executing `window.clearInterval(null or undefined)` without `if` statement wont cause any errors

fromjkeee16 (approved):
nice

3 small comments

src/App.tsx
ansofidev: Yes, thank you, but maybe timerId: number | undefined is incorrect because setInterval always returns a number, not undefined. Thanks for brining this to my attention, maybe better to use ReturnType<typeof setInterval> | null for accurate typing.

alexander-ignatow (approved):
👍  good job

PR #2752, opened 2025-03-22
solution

kami2693:
[DEMO LINK](https://kami2693.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,6 +4,7 @@ import { Clock } from './components/Clock';
 
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
+
   return `Clock-${value}`;
 }
 
@@ -25,8 +26,7 @@ export class App extends Component<{}, State> {
     document.addEventListener('click', this.handleClick);
 
     this.nameUpdateTimer = window.setInterval(() => {
-      const newName = getRandomName();
-      this.setState({ clockName: newName });
+      this.setState({ clockName: getRandomName() });
     }, 3300);
   }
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,37 +1,62 @@
...
+    clockName: 'Clock-0',
+  };
+
+  componentDidMount() {
+    document.addEventListener('contextmenu', this.handleContextMenu);
+    document.addEventListener('click', this.handleClick);
+
+    this.nameUpdateTimer = window.setInterval(() => {
+      const newName = getRandomName();
+      this.setState({ clockName: newName });
```
alexander-ignatow: Why have you used this intermediate variable?

```suggestion
      this.setState({ clockName: getRandomName() });
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface Props {
+  clockName: string;
+}
+
+export class Clock extends Component<Props> {
+  private timerId: number | null = null;
+
+  componentDidMount() {
+    this.timerId = window.setInterval(() => {
+      this.forceUpdate();
```
alexander-ignatow: `forceUpdate` is always a bad idea

Polease remove it and do the following:
1. Add current time as a state
2. Remove `forceUpdate` call and replace it with a state seter, with current time calculateion
3. remove current time calculation from render method and just display current state every render

alexander-ignatow (changes requested):
Looks good, but some improvements are needed to bring your code to it's fullest perfection :)

alexander-ignatow (approved):
Now looks good!

PR #2766, opened 2025-03-25
add clock

MariiaKapustkina:
[DEMO LINK](https://MariiaKapustkina.github.io/react_clock/)

### DIFF
--- a/src/Clock.tsx
+++ b/src/Clock.tsx
@@ -1,0 +1,85 @@
+import React from 'react';
+
+type State = {
+  today: Date;
+  clockName: string;
+};
+
+type Props = {
+  name: string;
+  onNameChange: (name: string) => void;
+};
+
+function getRandomName(): string {
+  const value = Date.now().toString().slice(-4);
+
+  return `Clock-${value}`;
+}
+
+export class Clock extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    today: new Date(),
+    clockName: this.props.name,
+  };
+
+  timerId: number | null = null;
+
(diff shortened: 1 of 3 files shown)


maxbilfed (approved):
Well done

src/Clock.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export class Clock extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    today: new Date(),
+  };
+
+  timerId: number | null = null;
+
+  nameTimeoutId: number | null = null;
```
alexander-ignatow, line 23: [minor] - variable names are not consistent:
I'd suggest either `timerId` and `nameTimerId` or `timeoutId` and `nameTimeoutId`

src/Clock.tsx
```diff
@@ -0,0 +1,70 @@
...
+  updateName = () => {
+    const newName = getRandomName();
+    const oldName = this.props.name;
+
+    // eslint-disable-next-line no-console
+    console.warn(`Renamed from ${oldName} to ${newName}`);
+    this.props.onNameChange(newName);
+
+    this.nameTimeoutId = window.setTimeout(this.updateName, 3300);
+  };
```
alexander-ignatow, line 34: You store the clock name in the App (that is ok) but for some reason update it from inside the Clock

That looks weird - it is better to move the name updating logic to the App and remove  `onNameChange` prop from the clock

alexander-ignatow (approved):
Good, but check out the some minor issues for the future

PR #2768, opened 2025-03-26
made clock

maxbilfed:
[DEMO LINK](https://maxbilfed.github.io/react_clock/)


src/App.tsx
```diff
@@ -1,37 +1,58 @@
...
+      event.preventDefault();
+      this.setState({ hasClock: false });
+    });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => {
+      this.setState({ hasClock: true });
+    });
```
alexander-ignatow: These listeners are never removed

src/Clock.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+type State = {
+  clockName: string;
+  time: string;
+};
+
+export class Clock extends React.Component<Props, State> {
+  state = {
+    clockName: 'Clock-0',
```
alexander-ignatow: You have the `clockName` state in the App and already pass it as a property to the Clock
Why do you need another clockName state here?

alexander-ignatow (changes requested):
Good job, but there are some issues

dothearts (approved):
Good job. 
Approved. 
Many thanks!)

PR #2795, opened 2025-04-16
add task solution

LadyLiza:


LadyLiza:
Just to clarify – the `Clock` component is completely unmounted when `hasClock` is false, since it’s conditionally rendered like this:

Because of that, the `componentWillUnmount` in `Clock.tsx` runs, and it clears the interval that logs the time.

So the `console.log` only happens when the Clock is visible.

PR #2938, opened 2025-09-22
add solution

George-Buzhenitsa:
https://George-Buzhenitsa.github.io/react_clock/


src/App.tsx
```diff
@@ -1,37 +1,60 @@
...
 function getRandomName(): string {
   const value = Date.now().toString().slice(-4);
 
   return `Clock-${value}`;
 }
 
-export const App: React.FC = () => {
-  const today = new Date();
-  let clockName = 'Clock-0';
+interface Props {};
```
vitalii-fedusov, line 11: why you keep empty interface?
you can accept clockName as a parameter or just pass empty object React.PureComponent<{}, State>

vitalii-fedusov (approved):
Looks good

src/App.tsx
George-Buzhenitsa: Thanks for usefull point))) I will take care of this in the future projects

PR #2955, opened 2025-10-21
Solution

Maevalar:
[DEMO LINK](https://Maevalar.github.io/react_clock/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,33 +21,30 @@ export class App extends React.Component {
 
   timerId = 1;
 
+  handleRightClick = (event: MouseEvent) => {
+    event.preventDefault(); // not to show the context menu
+
+    this.setState({ hasClock: false });
+  };
+
+  handleLeftClick = () => this.setState({ hasClock: true });
+
   componentDidMount(): void {
     this.timerId = window.setInterval(() => {
       this.setState({ clockName: getRandomName() });
       // eslint-disable-next-line no-console
     }, 3300);
 
-    document.addEventListener('contextmenu', (event: MouseEvent) => {
-      event.preventDefault(); // not to show the context menu
+    document.addEventListener('contextmenu', this.handleRightClick);
 
-      this.setState({ hasClock: false });
-    });
-
-    document.addEventListener('click', () => this.setState({ hasClock: true }));
-  }
-
-  componentDidUpdate(prevProps: {}, prevState: { clockName: string }): void {
-    if (this.state.hasClock && prevState.clockName !== this.state.clockName) {
-      // eslint-disable-next-line no-console
-      console.warn(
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,37 +1,64 @@
...
 
-  // this code stops the timer
-  window.clearInterval(timerId);
+  timerId = 1;
+
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ clockName: getRandomName() });
+      // eslint-disable-next-line no-console
+    }, 3300);
```
Aiiyuu, line 36: You should move `3300` into a dedicated constant, for example:

```javascript
const RANDOM_NAME_INTERVAL = 3300;
```

src/App.tsx
```diff
@@ -1,37 +1,64 @@
...
+    document.addEventListener('contextmenu', (event: MouseEvent) => {
+      event.preventDefault(); // not to show the context menu
+
+      this.setState({ hasClock: false });
+    });
 
-  return (
-    <div className="App">
-      <h1>React clock</h1>
+    document.addEventListener('click', () => this.setState({ hasClock: true }));
```
Aiiyuu: Don't you need to remove the event listener in **componentWillUnmount?** This violates checklist point 1.

[CODE KNOWLEDGE] - Check if you remove your event listener, when component will unmount from page.

BAD EXAMPLE:
```javascript
componentDidMount() {
  document.addEventListener('click', this.handleLeftClick);
}
```

GOOD EXAMPLE:
```javascript
componentDidMount() {
  document.addEventListener('click', this.handleLeftClick);
}

componentWillUnmount() {
  document.removeEventListener('click', this.handleLeftClick);
}
```

src/components/Clock.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export class Clock extends React.Component<Props> {
+  state: State = {
+    today: new Date(),
+  };
+
+  timerId = 0;
+
+  // This code starts a timer
```
Aiiyuu, line 18: Just write code that explains itself. In this case, you don't need to write such comments because it is obvious what your code is doing

src/components/Clock.tsx
```diff
@@ -0,0 +1,49 @@
...
+  componentDidMount(): void {
+    this.timerId = window.setInterval(() => {
+      this.setState({ today: new Date() }, () => {
+        // eslint-disable-next-line no-console
+        console.log(this.state.today.toUTCString().slice(-12, -4));
```
Aiiyuu, line 23: You’re using this piece of code twice:

```javascript
today.toUTCString().slice(-12, -4)
```

Maybe you should create a dedicated helper function for this?

Aiiyuu (approved):
**Overall, good job.** However, according to the checklist, you need to remove the event listeners when component will unmount from page

PR #2958, opened 2025-10-26
att1, fll works but 1 test fell, i dkw why

MaksimShe:
[DEMO LINK](https://MaksimShe.github.io/react_clock/)


nazkov (approved):
I didnt do this task, but I think its a good job ;)

